PR #3, opened 2018-03-12
Integrate delegates in adam's module approach 

satyamakgec:


adamdossa (approved):
Looks good. There are a couple of small bugs to fix on my side, but I'll do this after the merge.

PR #42, opened 2018-03-29
generate of ST/STO by polymath on the behalf of issuer

satyamakgec:


pabloruiz55 (changes requested):
Thanks @satyamakgec could you please move this test to another file? I'd like to keep this test separate to check on the steps required.
Also, make sure you review Adam's latest fixes to CappedSTO test so this works properly.

pabloruiz55 (approved):

PR #79, opened 2018-04-11
change alignment in migration script

satyamakgec:


adamdossa (approved):

pabloruiz55:
@satyamakgec can we add comments back into each step for the sake of documenting? So anyone coming to this gets familiar with what each operation does. Thanks!

PR #82, opened 2018-04-17
WIP - Oz adapter CLI

satyamakgec:
run - `node demo/adapter/generateSTO.js`
It only generates the generalize STO. If you both are good with this then I will create the dynamic factory creation as well.


pabloruiz55:
@satyamakgec are we ok with closing this? We decided it makes little sense to try to do this and we have diverged to much here.

satyamakgec:
Yes, It is better to close them

PR #99, opened 2018-05-08
burn securities

satyamakgec:


contracts/tokens/SecurityToken.sol
```diff
@@ -53,10 +57,12 @@ contract SecurityToken is ISecurityToken, StandardToken, DetailedERC20 {
         uint256 _timestamp
     );
 
+    event LogUpdateTokenDetails(bytes32 _oldDetails, bytes32 _newDetails);
     event LogGranularityChanged(uint256 _oldGranularity, uint256 _newGranularity);
     event LogModuleRemoved(uint8 indexed _type, address _module, uint256 _timestamp);
     event LogModuleBudgetChanged(uint8 indexed _moduleType, address _module, uint256 _budget);
     event Mint(address indexed to, uint256 amount);
+    event Burn(address indexed _burner, uint256 _value);
```
adamdossa: Can we change event names to:
Minted
Burnt
this keeps naming consistent (past tense) and differentiates a bit from the same function names.

contracts/tokens/SecurityToken.sol
```diff
@@ -302,4 +317,21 @@ contract SecurityToken is ISecurityToken, StandardToken, DetailedERC20 {
...
         }
     }
+
+    function setTokenBurner(address _tokenBurner) public onlyOwner {
+        tokenBurner = TokenBurner(_tokenBurner);
+    }
+
+    function burn(uint256 _value) checkGranularity(_value) public {
+        require(tokenBurner != address(0), "Token Burner contract address is not set yet");
+        require(_value <= balances[msg.sender], "Value should no be greater than the balance of meg.sender");
```
adamdossa: msg.sender rather than meg.sender

contracts/tokens/SecurityToken.sol
adamdossa: Maybe also move these events into IST20.sol?

pabloruiz55 (approved):

PR #123, opened 2018-06-01
Checkpoint Modules

adamdossa:


contracts/tokens/SecurityToken.sol
```diff
@@ -426,4 +499,66 @@ contract SecurityToken is ISecurityToken {
             sig = bytes4(uint(sig) + uint(_data[i]) * (2 ** (8 * (len - 1 - i))));
         }
     }
+
+    /**
+     * @dev Creates a checkpoint that can be used to query historical balances / totalSuppy
+     */
+    function createCheckpoint() public onlyModule(CHECKPOINT_KEY, true) {
+        require(currentCheckpointId < 2**256 - 1);
+        currentCheckpointId = currentCheckpointId + 1;
```
pabloruiz55: Why the above require instead of using SafeMath here?
currentCheckpointId = currentCheckpointId.add(1);

contracts/tokens/SecurityToken.sol
```diff
@@ -355,6 +426,7 @@ contract SecurityToken is ISecurityToken {
     function mint(address _investor, uint256 _amount) public onlyModule(STO_KEY, true) checkGranularity(_amount) returns (bool success) {
         adjustInvestorCount(address(0), _investor, _amount);
         require(verifyTransfer(address(0), _investor, _amount), "Transfer is not valid");
+        adjustTotalSupplyCheckpoints();
         totalSupply_ = totalSupply_.add(_amount);
         balances[_investor] = balances[_investor].add(_amount);
         emit Minted(_investor, _amount);
```
pabloruiz55: Shouldn't `adjustBalanceCheckpoints` be called here as well? If not an account would have a different amount of tokens than what balanceOfAt will return.

contracts/tokens/SecurityToken.sol
```diff
@@ -402,13 +474,14 @@ contract SecurityToken is ISecurityToken {
...
      * @dev Burn function used to burn the securityToken
-     * @param _value No. of token that get burned     
+     * @param _value No. of token that get burned
      */
     function burn(uint256 _value) checkGranularity(_value) public {
         adjustInvestorCount(msg.sender, address(0), _value);
         require(tokenBurner != address(0), "Token Burner contract address is not set yet");
         require(verifyTransfer(msg.sender, address(0), _value), "Transfer is not valid");
         require(_value <= balances[msg.sender], "Value should no be greater than the balance of msg.sender");
+        adjustTotalSupplyCheckpoints();
```
pabloruiz55: Shouldn't `adjustBalanceCheckpoints` be called here as well (below the burning of tokens)? If not an account would have a different amount of tokens than what balanceOfAt will return.

contracts/tokens/SecurityToken.sol
```diff
@@ -299,12 +310,70 @@ contract SecurityToken is ISecurityToken {
...
+            checkpointBalances[_investor].push(
+                Checkpoint({
+                    checkpointId: currentCheckpointId,
+                    value: balanceOf(_investor)
+                })
+            );
+            return;
+        }
+        //No new checkpoints since last update
+        if (checkpointBalances[_investor][checkpointBalances[_investor].length - 1].checkpointId == currentCheckpointId) {
```
pabloruiz55, line 357: This line only allows a SINGLE checkpointBalance to be created per checkpoint. If on checkpoint 1 account 0xabc makes 2 transfers, only the first one will be "recorded". Is that correct?

pabloruiz55 (approved):

pabloruiz55 (changes requested):
Left comments in previous action. Approved by mistake.

contracts/tokens/SecurityToken.sol
adamdossa: That's correct - the first time a transfer is made after a given checkpoint, the balance *before* the transfer is written out for the checkpoint, so that the balance written out corresponds to the time at which the checkpoint was made.

contracts/tokens/SecurityToken.sol
adamdossa: Agreed - added.

contracts/tokens/SecurityToken.sol
adamdossa: Agreed - added.

contracts/tokens/SecurityToken.sol
adamdossa: Yep - SafeMath would work here. My thinking is that we may want to limit the number of checkpoints more than the technical upper limit, but I need to do some gas limit calculations to check what a sensible limit is (it should be in the 1000s but need to do the analysis).

adamdossa:
FYI - I improved the fuzz testing so that it catches the minting issue above for future changes.

pabloruiz55 (approved):

PR #127, opened 2018-06-06
finishMinting() added

satyamakgec:


contracts/tokens/SecurityToken.sol
```diff
@@ -89,6 +93,7 @@ contract SecurityToken is ISecurityToken {
         }
         if (_fallback && !isModuleType) {
             require(msg.sender == owner, "Sender is not owner");
+            require(!mintingFinished);
```
adamdossa: Shouldn't this require statement just be in the mint function?

adamdossa (commented):
Not sure we should be checking msg.sender == owner before using finishedMinting in the mint function. My feeling is that if the token is set to finishedMinting no further tokens should be minted by anyone (including STOs).

pabloruiz55 (approved):

PR #138, opened 2018-06-13
Upgradability-feature

satyamakgec:


contracts/interfaces/IRegistry.sol
```diff
@@ -2,13 +2,58 @@ pragma solidity ^0.4.23;
...
+    /**
+     * @dev change the contract address
+     * @param _nameKey is the key for the contract address mapping
+     * @param _newAddress is the new contract address
+     */
+    function changeAddress(string _nameKey, address _newAddress) public onlyOwner {
+        address oldAddress = getAddress(_nameKey);
+        storedAddresses[keccak256(_nameKey)] = _newAddress;
+        emit LogChangeAddress(_nameKey, oldAddress, _newAddress);
+    }
```
adamdossa: Can we merge setAddress & changeAddress (so we just have a single changeAddress function which is used both for initialisation and subsequent changes).

contracts/interfaces/IRegistry.sol
```diff
@@ -2,13 +2,58 @@ pragma solidity ^0.4.23;
...
-contract IRegistry is Ownable {
+contract IRegistry is Ownable, IPausable {
+
+    /*
+    Valid Address Keys
+        TR_Address -> getAddress("TR_Address")
+        STR_Address -> getAddress("STR_Address")
+        MR_Address -> getAddress("MR_Address")
+        POLY_Address -> getAddress("POLY_Address")
+        */
```
adamdossa: Could we use a more consistent naming scheme for all of these variables and keys.

Variables should be tickerRegistry / moduleRegistry / securityTokenRegistry and keys should be "TickerRegistry" / "ModuleRegistry" / "SecurityTokenRegistry" - please could you change this in all of the contracts - generally prefer to avoid "_" in variable names, and instead use "_" to indicate a parameter which is passed into a function.

contracts/tokens/SecurityToken.sol
```diff
@@ -115,22 +117,22 @@ contract SecurityToken is ISecurityToken {
...
+        address _STR_Address
     )
     public
     DetailedERC20(_name, _symbol, _decimals)
     {
         //When it is created, the owner is the STR
-        moduleRegistry = ISecurityTokenRegistry(_securityTokenRegistry).moduleRegistry();
-        polyToken = ERC20(ISecurityTokenRegistry(_securityTokenRegistry).polyAddress());
+        STR_Address = _STR_Address;
+        polyToken = ERC20(IRegistry(_STR_Address).getAddress("POLY_Address"));
```
adamdossa: Rather than storing `polyToken` can we just look it up via the IRegistry(_STR_Address).getAdress("POLY_Address")) when we need it?

contracts/interfaces/IRegistry.sol
adamdossa: And the same for POLY_Address - could we use polyToken everywhere.

contracts/tokens/SecurityToken.sol
thegostep: What is the advantage in doing so? Is it because of the gas difference between memory vs storage?

thegostep:
Adam's suggestions are implemented

adamdossa (approved):

PR #155, opened 2018-06-25
Voting module

thegostep:
A voting module for single-issue voting


adamdossa (approved):

pabloruiz55 (changes requested):
"Requesting changes" only to block merging until after mainnet release

PR #157, opened 2018-06-27
 minting can be forbids at STO level or at Issuer level independently

satyamakgec:


pabloruiz55 (changes requested):
Looks good @satyamakgec 
Need you to fix the CLIs too as they won't work anymore with this breaking change.

pabloruiz55 (approved):

PR #160, opened 2018-06-29
Cli gas estimate

VictorVicente:


satyamakgec (changes requested):
Getting a revert error! when muliMint script run at the time of creation of ST20Generator
`Press "Y" if you have a list of affiliates addresses with you otherwise hit Enter and get the minted tokens to a particular address: Y`.

pabloruiz55:
This is partly my fault since I changed one address in the whitelist csv and not the mint, so it's basically trying to mint tokens to a non-whitelisted address, reverting and making the whole script exit.

As Victor indicated, this happens on master too. He's going to fix it by catching the transfer error before attempting to blindly transfer.

pabloruiz55 (approved):

satyamakgec (approved):

PR #161, opened 2018-06-30
Add test for concurrent STOs

thegostep:
In capped_sto I cleaned up some global variables to prevent overwriting which was causing the new tests to fail.
I made concurrent_sto to test adding and investing in multiple different types of STOs. It can easily be extended to cover new STO modules.

### DIFF
--- a/test/capped_sto.js
+++ b/test/capped_sto.js
@@ -51,10 +51,8 @@ contract('CappedSTO', accounts => {
     let I_STVersion;
     let I_SecurityToken_ETH;
     let I_SecurityToken_POLY;
-    let I_CappedSTO_ETH1;
-    let I_CappedSTO_ETH2;
-    let I_CappedSTO_POLY1;
-    let I_CappedSTO_POLY2;
+    let I_CappedSTO_Array_ETH = [];
+    let I_CappedSTO_Array_POLY = [];
     let I_PolyToken;
 
     // SecurityToken Details for funds raise Type ETH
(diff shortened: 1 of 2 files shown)


satyamakgec (dismissed):
Current PR is working fine. Could you please add one more test case which will loop 20 times to hit the limit of attaching the STO where you can attach different STO as the dummySTO, cappedSTO, pre-saleSTO etc. same for investment in those STO. This way we will 100% sure with the test case and as well as with our solidity code .:wink:

satyamakgec (commented):
Current PR is working fine. Could you please add one more test case which will loop 20 times to hit the limit of attaching the STO where you can attach different STO as the dummySTO, cappedSTO, pre-saleSTO etc. same for investment in those STO. This way we will 100% sure with the test case and as well as with our solidity code .:wink:

test/capped_sto.js
```diff
@@ -665,22 +673,149 @@ contract('CappedSTO', accounts => {
...
+
+        it("Should successfully attach STO modules up to the limit and fail at the limit", async () => {
+            let startTime = latestTime() + duration.days(1);
+            let endTime = startTime + duration.days(30);
+
+            await I_PolyToken.getTokens(cappedSTOSetupCost*19, token_owner);
+            await I_PolyToken.transfer(I_SecurityToken_ETH.address, cappedSTOSetupCost*19, { from: token_owner});
+            let bytesSTO = web3.eth.abi.encodeFunctionCall(functionSignature, [startTime, endTime, cap, rate, fundRaiseType, account_fundsReceiver]);
+
+            for (var STOIndex = 2; STOIndex < 20; STOIndex++) {
```
pabloruiz55: SecurityToken has this var: `uint8 public constant MAX_MODULES = 20;`
instead of hardcoding 20 in the for loop use that one since MAX_MODULES might change in the future.

test/capped_sto.js
```diff
@@ -665,22 +673,149 @@ contract('CappedSTO', accounts => {
...
+                console.log(`         tx revert -> reached cap number of modules attached`.grey);
+                ensureException(error);
+                errorThrown = true;
+            }
+            assert.ok(errorThrown, message);
+        });
+
+        it("Should successfully invest in all STO modules attached", async () => {
+            await increaseTime(duration.days(2));
+            for (var STOIndex = 2; STOIndex < 20; STOIndex++) {
```
pabloruiz55: Same as above: use `uint8 public constant MAX_MODULES = 20;`

pabloruiz55 (approved):

PR #176, opened 2018-07-26
 Prevent upgrade from impacting ability to attach modules

thegostep:
By passing the address of the STR from the security token, we are allowing the issuer to continue attaching existing registered modules despite registry upgrade.

### DIFF
--- a/contracts/ModuleRegistry.sol
+++ b/contracts/ModuleRegistry.sol
@@ -40,6 +40,7 @@ contract ModuleRegistry is IModuleRegistry, Pausable, RegistryUpdater, ReclaimTo
    /**
     * @notice Called by a security token to notify the registry it is using a module
     * @param _moduleFactory is the address of the relevant module factory
+    * @param _securityTokenRegistry is the address of the Security Token Registry
     */
     function useModule(address _moduleFactory, address _securityTokenRegistry) external {
         //If caller is a registered security token, then register module usage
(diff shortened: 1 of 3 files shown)


pabloruiz55:
@adamdossa please review and indicate if this is a sensible fix we should be pushing out now or push to after mainnet.

@thegostep test cases / CLI need to be adapted.

thegostep:
@pabloruiz55 The tests and CLI are not impacted by this fix since`useModule` is never called directly by a user

adamdossa:
I guess it's a bit pointless checking if the caller is a ST, when the caller can pass in the STR which is used to make the check (i.e. a malicious caller could pass in a dummy STR that always returns true from `isSecurityToken`) - or am I missing something?

Another possible approach might be to split out the data holding a list of PM tokens into a separate very simple contract, which can't be modified / upgraded (similar to something like PolymathRegistry) - i.e. just a list of PM STs and nothing else. At the moment there is always a reasonable chance we'll end up upgrading the STR because it has quite a lot of business logic.

In the current version, I think it is reasonable that if Polymath goes (very) rogue, then issuers won't be able to add new modules. Ultimately we can't impact their current modules or tokens, just stop them adding new features. In future versions we can keep improving all of this stuff (splitting out data & logic, making logic upgradable, and data not so much).

adamdossa:
Closing as no longer needed.

PR #188, opened 2018-08-16
Ticker registry revamping

satyamakgec:
Completed tasks : 
1. Add getter function to get the list of tickers hold by the issuer #189 .
2. `transferTickerOwnership()` function added #191 .
3. `expiryTimestamp` is added #192 .
4. `addCustomTicker()` is added #190.


adamdossa:
@satyamakgec One question on the:
`_renounceTickerOwnership`
function. If a single address registered lots and lots of tickers, then it may become impossible to call this function due to block gas limits.

It may be better to hold a separate mapping from ticker > index into the owners array of tickers so that we can look up in constant time which index in the `tokensOwnedByUser` array we need to remove (can be removed by replacing with the last entry in the list, and then reducing length by one)

satyamakgec:
Amend the changes @adamdossa

contracts/TickerRegistry.sol
```diff
@@ -23,19 +23,33 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
     // Details of the symbol that get registered with the polymath platform
     struct SymbolDetails {
         address owner;
-        uint256 timestamp;
+        uint256 registeredTimestamp;
```
pabloruiz55: Would rename to `registrationDate`

contracts/TickerRegistry.sol
```diff
@@ -23,19 +23,33 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
...
+        uint256 expiredTimestamp;
```
pabloruiz55: would rename to `expiryDate`

contracts/TickerRegistry.sol
```diff
@@ -91,11 +107,29 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
...
+   
+    /**
+     * @notice Register the ticker without paying the fee 
+       Once the token symbol is registered to its owner then no other issuer can claim
+       Its ownership. If the symbol expires and its issuer hasn't used it, then someone else can take it.
+     * @param _symbol token symbol
+     * @param _tokenName Name of the token
+     * @param _swarmHash Off-chain details of the issuer and token
+     */
+    function addCustomTicker(string _symbol, string _tokenName, bytes32 _swarmHash) public onlyOwner {
```
pabloruiz55: addCustomTicker should allow us to fully recreate the ticker if needed, so probably it should accept every single parameter as it comes.

contracts/TickerRegistry.sol
```diff
@@ -170,6 +181,80 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
...
+
+    /**
+     * @notice Transfer the ownership of the ticker
+     * @dev _newOwner Address whom ownership to transfer
+     * @dev _ticker Symbol 
+     */
+    function transferTickerOwnership(address _newOwner, string _ticker) public whenNotPaused {
+        string memory ticker = upper(_ticker);
+        require(_newOwner != address(0), "Address should not be 0x");
+        require(bytes(ticker).length > 0, "Ticker length should be greator than 0");
```
pabloruiz55: typo => greater

contracts/TickerRegistry.sol
```diff
@@ -170,6 +181,80 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
...
+        tokensOwnedByUser[_newOwner].push(stringToBytes32(ticker));
+        tickerIndex[ticker] = tokensOwnedByUser[_newOwner].length - 1;
+        emit LogChangeTickerOwnership(ticker, msg.sender, _newOwner);
+    }
+
+    /**
+     * @notice Use to get the ticker list as per the owner
+     * @param _owner Address which owns the list of tickers 
+     */
+    function getTickerOwnedByUser(address _owner) public view returns(bytes32[]) {
```
pabloruiz55: rename to `getTickersByOwner`

contracts/TickerRegistry.sol
```diff
@@ -92,11 +108,33 @@ contract TickerRegistry is ITickerRegistry, Util, Pausable, RegistryUpdater, Rec
...
+     * @notice Register the ticker without paying the fee 
+       Once the token symbol is registered to its owner then no other issuer can claim
+       Its ownership. If the symbol expires and its issuer hasn't used it, then someone else can take it.
+     * @param _symbol token symbol
+     * @param _tokenName Name of the token
+     * @param _swarmHash Off-chain details of the issuer and token
+     * @param _registrationDate Date on which ticker get registered
+     * @param _expiryDate Expiry date of the ticker
+     */
+    function addCustomTicker(string _symbol, string _tokenName, bytes32 _swarmHash, uint256 _registrationDate, uint256 _expiryDate) public onlyOwner {
```
pabloruiz55: `addCustomTicker` would be used to register a ticker on behalf of someone else, so one parameter should be `_owner` and not use msg.sender

satyamakgec: Do we think to use that function to migrate the tickers? I have thought a more generalize way where `polymathAccount` add the ticker then transfer the ownership of the ticker.

pabloruiz55: I don't see the point in registering them to Polymath and then transferring ownership, vs straight register them to someone else. This would have come in handy for the Polycon regs we did, so we didn't need to send POLY alongside. And also, very useful to migrate data later on.

pabloruiz55 (approved):

PR #193, opened 2018-08-21
minor fixes

### LINKED ISSUE #196: Standardize STOs so we can ensure all STO modules can have the same data fetched

satyamakgec:
fix #196 #195 #194


pabloruiz55:
Looks good. Though I'd rather have a 2nd pair of eyes here. @adamdossa

adamdossa:
@satyamakgec for:
```
getReputationOfFactory
getModuleListOfType
```
don't we already have identical public getters in the ABI as these correspond to the public state variables:
```
mapping (address => address[]) public reputation;
mapping (uint8 => address[]) public moduleList;
```
?

contracts/modules/PermissionManager/GeneralPermissionManagerFactory.sol
```diff
@@ -26,7 +26,7 @@ contract GeneralPermissionManagerFactory is IModuleFactory {
         if(setupCost > 0)
             require(polyToken.transferFrom(msg.sender, owner, setupCost), "Failed transferFrom because of sufficent Allowance is not provided");
         address permissionManager = new GeneralPermissionManager(msg.sender, address(polyToken));
-        emit LogGenerateModuleFromFactory(address(permissionManager), getName(), address(this), msg.sender, now);
+        emit LogGenerateModuleFromFactory(address(permissionManager), getName(), address(this), msg.sender, setupCost, now);
         return address(permissionManager);
```
adamdossa: Should just be:
`return permissionManager;`

adamdossa: or:
`address permissionManager = new ...`
should be:
PermissionManager permissionManager = new ...`

satyamakgec: Yes it could be only `permissionManager`. and for the getter functions it is used for CLI and stats scripts #194. I think can use directly the public functions but @VictorVicente was facing some problem regarding it.

contracts/modules/STO/CappedSTO.sol
```diff
@@ -178,9 +161,9 @@ contract CappedSTO is ISTO, ReentrancyGuard {
             endTime,
             cap,
             rate,
-            fundsRaised,
+            (fundRaiseType[uint8(FundRaiseType.POLY)]) ? fundsRaisedPOLY: fundsRaisedPOLY,
```
pabloruiz55: This looks wrong if -else is returning same value

satyamakgec: resolved

adamdossa (approved):

PR #202, opened 2018-08-22
Add registration timestamp

satyamakgec:


contracts/SecurityTokenRegistry.sol
```diff
@@ -81,7 +81,7 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, Util, Pausable, Regist
         require(_owner != address(0));
         require(!(ITickerRegistry(tickerRegistry).isReserved(symbol, _owner, _name, _swarmHash)), "Trying to use non-valid symbol");
         symbols[symbol] = _securityToken;
-        securityTokens[_securityToken] = SecurityTokenData(symbol, _tokenDetails);
+        securityTokens[_securityToken] = SecurityTokenData(symbol, _tokenDetails, now);
```
pabloruiz55: Instead of passing `now`, the date should be a parameter, so if we need to migrate the data we can replicate the old registration date

contracts/SecurityTokenRegistry.sol
```diff
@@ -81,7 +81,7 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, Util, Pausable, Regist
...
         emit LogAddCustomSecurityToken(_name, symbol, _securityToken, now);
```
pabloruiz55: same here, instead of `now` that should be an extra parameter being passed so we can migrate the reg date

satyamakgec:
Don't merge it I am creating a new one as a `development-1.5.0` base branch

PR #203, opened 2018-08-22
fix #200

### LINKED ISSUE #200: Make it possible to change a ModuleFactory's title, name ,description

satyamakgec:


pabloruiz55:
Looks good, let's use this opportunity to add another property: version, just in case we need it later one.

satyamakgec:
added `version` variable

pabloruiz55 (approved):

PR #205, opened 2018-08-22
fix #198

### LINKED ISSUE #198: Review the getModuleByName logic
Names for modules are not unique. We need to figure out a different way of doing this or at least enforcing module name uniqueness

satyamakgec:
For now, It returns the array of `address` and `bytes32`.  In future or when we do upgradeability stuff for modules then we can also attach a unique parameter to a module as `tranche` for securityToken, Like a standard for Modules that will be the best approach to make the uniqueness in the ecosystem by not forcing the name uniqueness.


pabloruiz55:
We might want to add this new function instead of replacing the old one in order to keep backwards compatibility.

pabloruiz55 (approved):

PR #224, opened 2018-08-28
Freeze Minting

thegostep:
### Requirements

- [x] Issuer MUST be able to call a function which prevents them from minting tokens in the future
- [x] This issue MUST be disabled by default until enabled by Polymath
- [x] Function MUST emit event notifying minting is done
- [x] Issuer MUST be able to mint tokens at any time before calling this function
- [x] Issuer MUST NOT be able to mint tokens at any time after calling this function

### Changes
*ISecurityToken.sol*
* remove function `finishMintingIssuer()` and function `finishMintingSTO()`
* add function `freezeMinting()`

*SecurityToken.sol*
* rename bool `freeze` to bool `transfersFrozen`
* add bool `mintingFrozen`
* remove bool `finishedIssuerMinting` and bool `finishedSTOMinting`
* remove event `LogFinishMintingSTO` and event `LogFinishMintingIssuer`
* add event `LogFreezeMinting`
* remove function `finishMintingIssuer()` and function `finishMintingSTO()`
* add function `freezeMinting()`
* change modifier `isMintingAllowed` to only check for `mintingFrozen`
* change modifier `onlyModule()` to remove fallback on `onlyOwner`
* add modifier `onlyModuleOrOwner()`

create *IFeatureRegisty.sol*
create *FeatureRegisty.sol*

*RegistryUpdater.sol*
* add `featureRegistry`


pabloruiz55:
What's the reasoning behind putting the Polymath allowFreezeMinting switch in the ModuleRegistry? Not that any of the other registries make more sense, though.
Perhaps what's more sensible to do is to create a new Polymath owned registry where all these regulatory feature switches go in, rather. @thegostep

contracts/ModuleRegistry.sol
```diff
@@ -118,6 +121,21 @@ contract ModuleRegistry is IModuleRegistry, Pausable, RegistryUpdater, ReclaimTo
         }
     }
 
+    /**
+     * @notice One-way feature switch to enable permanently freezing minting
```
pabloruiz55: Should our switch be one-way only?

contracts/tokens/SecurityToken.sol
```diff
@@ -49,8 +49,11 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
     // Reference to token burner contract
     ITokenBurner public tokenBurner;
 
-    // Use to halt all the transactions
-    bool public freeze = false;
+    // Use to temporarily halt all transactions
+    bool public transfersFrozen;
```
pabloruiz55, line 54: Weren't we going to differentiate permanent vs temporary flags with the naming?

contracts/tokens/SecurityToken.sol
```diff
@@ -515,29 +507,23 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-     * @dev Can only be called by the STO attached to the token (Or by the ST owner if there's no STO attached yet)
-     * @param _investor Address to whom the minted tokens will be dilivered
-     * @param _amount Number of tokens get minted
+     * @dev Can only be called by the issuer or STO attached to the token
+     * @param _investor Address where the minted tokens will be delivered
+     * @param _amount Number of tokens be minted
      * @return success
      */
-    function mint(address _investor, uint256 _amount) public onlyModule(STO_KEY, true) checkGranularity(_amount) isMintingAllowed() returns (bool success) {
+    function mint(address _investor, uint256 _amount) public onlyModule(STO_KEY, true) checkGranularity(_amount) mintingAllowed() returns (bool success) {
```
pabloruiz55: Personal preference, but I like it better when modifiers say isSomething, hasSomething, onlySomething, gives you a better idea of how it's modifying the function.

migrations/2_deploy_contracts.js
```diff
@@ -213,24 +213,36 @@ module.exports = function (deployer, network, accounts) {
...
-      console.log('*** General Permission Manager Factory: ', GeneralPermissionManagerFactory.address, '***')
-      console.log('*** Count Transfer Manager Factory: ', CountTransferManagerFactory.address, '***')
-      console.log('*** Percentage Transfer Manager Factory: ', PercentageTransferManagerFactory.address, '***')
-      console.log('*** ETH Dividends Checkpoint Factory: ', EtherDividendCheckpointFactory.address, '***')
-      console.log('*** ERC20 Dividends Checkpoint Factory: ', ERC20DividendCheckpointFactory.address, '***')
-      console.log('-----------------------------------')
-      console.log('\n')
+      console.log('\n');
+      console.log(`
+      -------------------- Polymath Network Smart Contracts: --------------------
```
pabloruiz55: Ahhh Love this improvement!
Needs to have the ManualApproval TM address added to the deployment. This was done in 1.4.0 final commit.

contracts/tokens/SecurityToken.sol
pabloruiz55: There's a big change missing here. Mint has the onlyModule modifier, which still prevents minting from happening if there's an STO attached. onlyModule logic has to be reviewed. 
Tests should have caught this... Please review why the tests didn't and fix them.

contracts/ModuleRegistry.sol
thegostep: I don't have a strong opinion on this - will modify to allow turning off

contracts/tokens/SecurityToken.sol
thegostep: I left it as a separate task, keeping freeze for now makes it easier to come back to

contracts/tokens/SecurityToken.sol
thegostep: `whenMintingAllowed()` ?

migrations/2_deploy_contracts.js
thegostep: I am terrible at promise syntax. Maybe @satyamakgec can help here

contracts/tokens/SecurityToken.sol
thegostep: I believe this was a bug we decided not to fix. I definitely think the logic of onlyModule needs to be simplified. I've added a task to that effect

thegostep:
@pabloruiz55 The reason for putting the feature switch in the MR is because it is the only registry with a link to the deployed ST since they communicate for module deployment. WRT a new feature switch contract, it seems like a bit of an overkill and would create additional dependencies. I am still not familiar with the new upgradability pattern so not sure what will be the implications. Maybe @adamdossa has more thoughts.

contracts/ModuleRegistry.sol
pabloruiz55: ok. Yes, let's make it possible to have Polymath turn off/on

contracts/tokens/SecurityToken.sol
pabloruiz55: 👍

contracts/tokens/SecurityToken.sol
pabloruiz55: `isMintingAllowed()` sounds better.

migrations/2_deploy_contracts.js
pabloruiz55: No need to "code" what needs to be done here is to merge master into dev-1.5.0 since it's behind a few commits.

contracts/tokens/SecurityToken.sol
pabloruiz55: Current functionality is that minting is not possible while there's an STO attached. As per the task at hand we need to change it so onlyModule doesn't prevent minting when there's an STO.

pabloruiz55:
@thegostep WRT where this switch goes, let's keep it simple for now: 
1. Create new registry contract and move the logic added to the MR there.
2. Add that contract to the PolymathRegistry as a new key
3. From the SecurityToken contract you can access the new contract through the PolymathRegistry and get the flag.

contracts/FeatureRegistry.sol
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * @notice change a feature status
+     * @param _nameKey is the key for the feature status mapping
+     * @param _newStatus is the new feature status
+     */
+    function setFeatureStatus(string _nameKey, bool _newStatus) public onlyOwner {
+        bytes32 key = keccak256(bytes(_nameKey));
+        require(featureStatus[key] != _newStatus);
```
adamdossa: Can we add a revert error message here.

adamdossa (approved):

PR #244, opened 2018-09-03
Module Registry removeModule() function

thegostep:
Please double check logic for dealing with array index 0. My approach has been to set index 0 of the array to the 0x0 address and prevent from removing, the side effect is that getting the list of module factories will always include the 0x0 address.


pabloruiz55:
@thegostep why not just skip the whole re-ordering code block if array length is == 1?

contracts/ModuleRegistry.sol
```diff
@@ -20,6 +20,8 @@ contract ModuleRegistry is IModuleRegistry, Pausable, RegistryUpdater, ReclaimTo
     mapping (address => address[]) public reputation;
     // Mapping contain the list of addresses of Module factory for a particular type
     mapping (uint8 => address[]) public moduleList;
+    // Mapping to store the index of the moduleFacorty in the moduleList
```
pabloruiz55: typo => moduleFactory

contracts/ModuleRegistry.sol
```diff
@@ -67,14 +75,47 @@ contract ModuleRegistry is IModuleRegistry, Pausable, RegistryUpdater, ReclaimTo
     function registerModule(address _moduleFactory) external whenNotPaused returns(bool) {
         require(registry[_moduleFactory] == 0, "Module factory should not be pre-registered");
         IModuleFactory moduleFactory = IModuleFactory(_moduleFactory);
-        require(moduleFactory.getType() != 0, "Factory type should not equal to 0");
-        registry[_moduleFactory] = moduleFactory.getType();
-        moduleList[moduleFactory.getType()].push(_moduleFactory);
+        uint8 kind = moduleFactory.getType();
```
pabloruiz55: instead of "kind" use "type"

thegostep: type is a reserved keyword [documentation](https://solidity.readthedocs.io/en/v0.4.21/miscellaneous.html#reserved-keywords)

thegostep:
~~the check on line 99 does the same implicit check as 'array length is == 1'~~

edit: 
I miss-understood. What you are referring to is that there is no need to re-order if removing the last element in the array.

I can add a check `if (index == last)` and delete without re-ordering. I don't think it makes a big difference from a gas perspective.

contracts/ModuleRegistry.sol
pabloruiz55: I mean, moduleFactoryType or _type, don't want to be introducing different names.

pabloruiz55:
In response to this: 

> Please double check logic for dealing with array index 0. My approach has been to set index 0 of the array to the 0x0 address and prevent from removing, the side effect is that getting the list of module factories will always include the 0x0 address.

I meant instead of adding an item to index 0, change the logic so if there's only one item you don't re-order but just remove that item.

contracts/ModuleRegistry.sol
thegostep: okay fixing

pabloruiz55:
Looks good to me now. @adamdossa  please review.

pabloruiz55 (approved):

PR #259, opened 2018-09-16
Remove max modules

### LINKED ISSUE #260: Remove looping over module in onlyOwner, add archive / unarchive / remove modules
Remove restrictions on the number of modules by optimising data structures.

Introduce archiving / unarchiving modules and removing modules as separate concepts

adamdossa:
Fix #260 

@VictorVicente to make corresponding CLI changes

Uses an internal function for verifyTransfer rather than signature matching on `msg.data`

Removes MAX_MODULES and changes the `onlyModule` modifier to be constant time.

Modules can be:
archived: makes a module not valid for the purposes of `onlyModule`. Can be unarchived
unarchived: reverts archive
remove: permanently remove module (can't be re-added)

NB: We still have a dependency on the number of TM modules attached as we loop over these in `verifyTransfer`.

### DIFF
--- a/contracts/interfaces/ISecurityToken.sol
+++ b/contracts/interfaces/ISecurityToken.sol
@@ -55,7 +55,7 @@ interface ISecurityToken {
      * @return uint256 name index
 
      */
-    function getModule(address _module) external view;
+    function getModule(address _module) external view returns(bytes32, address, address, bool, uint8, uint256, uint256);
 
     /**
      * @notice returns module list for a module name
--- a/contracts/interfaces/ISecurityToken.sol
+++ b/contracts/interfaces/ISecurityToken.sol
@@ -44,27 +44,32 @@ interface ISecurityToken {
     function checkPermission(address _delegate, address _module, bytes32 _perm) external view returns (bool);
 
     /**
+     * @notice Returns module list for a module type
+     * @param _module address of the module
+     * @return bytes32 name
+     * @return address module address
+     * @return address module factory address
+     * @return bool module archived
+     * @return uint8 module type
+     * @return uint256 module index
+     * @return uint256 name index
+
+     */
+    function getModule(address _module) external view;
+
+    /**
+     * @notice returns module list for a module name
+     * @param _name name of the module
+     * @return address[] list of modules with this name
+     */
+    function getModulesByName(bytes32 _name) external view returns (address[]);
+
+    /**
      * @notice returns module list for a module type
-     * @param _moduleType is which type of module we are trying to access
-     * @param _moduleIndex is the index of the module within the chosen type
+     * @param _type type of the module
+     * @return address[] list of modules with this type
      */
-    function getModule(uint8 _moduleType, uint _moduleIndex) external view returns (bytes32, address);
-
-    /**
-     * @notice returns a module that matches the provided name - will return first match
-     * @param _moduleType is which type of module we are trying to access
-     * @param _name is the name of the module within the chosen type
-     */
-    function getModuleByName(uint8 _moduleType, bytes32 _name) external view returns (bytes32, address);
-
-    /**
-     * @notice returns all the modules that match the provided name within a module type
-     * @param _moduleType is the type of module we are trying to get
-     * @param _name is the name of the module within the chosen type
-     * @return bytes32
-     * @return address
-     */
-    function getAllModulesByName(uint8 _moduleType, bytes32 _name) external view returns (bytes32[], address[]);
+    function getModulesByType(uint8 _type) external view returns (address[]);
 
     /**
      * @notice Queries totalSupply at a specified checkpoint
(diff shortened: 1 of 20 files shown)


contracts/tokens/SecurityToken.sol
```diff
@@ -229,108 +235,113 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-                modules[_moduleType][_moduleIndex].name,
-                modules[_moduleType][_moduleIndex].moduleAddress
-            );
-        } else {
-            return ("", address(0));
+    * @notice Unarchives a module attached to the SecurityToken
+    * @param _module address of module to unarchive
+    */
+    function unarchiveModule(address _module) external onlyOwner {
+        require(modulesToData[_module].isArchived, "Already archived");
```
pabloruiz55: error string should be "Already un-archived"

contracts/tokens/SecurityToken.sol
```diff
@@ -229,108 +235,113 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
+        emit LogModuleUnarchived(modulesToData[_module].moduleType, _module, now);
+        modulesToData[_module].isArchived = false;
+    }
+
+    /**
+    * @notice Unarchives a module attached to the SecurityToken
```
pabloruiz55: Fix natspec comment

contracts/tokens/SecurityToken.sol
```diff
@@ -229,108 +235,113 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
+    * @param _module address of module to unarchive
+    */
+    function removeModule(address _module) external onlyOwner {
```
pabloruiz55: I'd require the module to be first archived before being able to remove it in order to prevent mistakes.

pabloruiz55:
Left a few small comments. Otherwise, looks good.
Only wondering if it we should really allow them to remove a module vs only just allowing them to archive them

contracts/interfaces/ISecurityToken.sol
```diff
@@ -44,27 +44,32 @@ interface ISecurityToken {
...
+     * @return address module address
+     * @return address module factory address
+     * @return bool module archived
+     * @return uint8 module type
+     * @return uint256 module index
+     * @return uint256 name index
+
      */
-    function getModule(uint8 _moduleType, uint _moduleIndex) external view returns (bytes32, address);
+    function getModule(address _module) external view;
```
satyamakgec: return parameter missing it should be 
`function getModule(address _module) external view returns(bytes32, address, address, bool, uint8, uint256, uint256)`

contracts/tokens/SecurityToken.sol
```diff
@@ -122,23 +133,23 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-    modifier onlyModule(uint8 _moduleType) {
-        bool isModuleType = false;
-        for (uint8 i = 0; i < modules[_moduleType].length; i++) {
-            isModuleType = isModuleType || (modules[_moduleType][i].moduleAddress == msg.sender);
-        }
-        require(isModuleType, "msg.sender is not correct module type");
+    modifier onlyModule(uint8 _type) {
+        require(modulesToData[msg.sender].module == msg.sender, "Address mismatch");
+        require(modulesToData[msg.sender].moduleType == _type, "Type mismatch");
+        require(!modulesToData[msg.sender].isArchived, "Module archived");
```
satyamakgec: can we replace all 3 require statements with `require(isModule(msg.sender, _type));`
and add a new function          

        function isModule(address _module, uint8 _type) internal view returns (bool) {
             // Add here all 3 require statements            
             return true;
        }

This will help us to reuse the code and saves the size as well

contracts/tokens/SecurityToken.sol
```diff
@@ -122,23 +133,23 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-        bool isModuleType = false;
-        for (uint8 i = 0; i < modules[_moduleType].length; i++) {
-            isModuleType = isModuleType || (modules[_moduleType][i].moduleAddress == msg.sender);
+    modifier onlyModuleOrOwner(uint8 _type) {
+        if (msg.sender == owner) {
+            _;
+        } else {
+          require(modulesToData[msg.sender].module == msg.sender, "Address mismatch");
+          require(modulesToData[msg.sender].moduleType == _type, "Type mismatch");
+          require(!modulesToData[msg.sender].isArchived, "Module archived");
```
satyamakgec: here we can also replace this with one require statement `require(isModule(msg.sender, _type));`

contracts/tokens/SecurityToken.sol
```diff
@@ -64,7 +64,12 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
 
     struct ModuleData {
         bytes32 name;
-        address moduleAddress;
+        address module;
+        address moduleFactory;
```
satyamakgec, line 68: is there any use to store the `moduleFactory` address here? or we have any future functionality depends on it

adamdossa: It's a good question. I was thinking we might want to look up fees / version etc. direct from the factory rather than storing locally in the ST. Didn't seem to do much harm to add this for now incase needed in the future.

adamdossa:
Thanks @satyamakgec @pabloruiz55 - have resolved comments.

pabloruiz55 (approved):

PR #262, opened 2018-09-18
Single Trade Volume Restriction

### LINKED ISSUE #258: Single-Trade Volume Restriction Module Development Task
# Single-Trade Volume Restriction Module 

Module Type | Bounty | Completion deadline
-- | -- | --
Transfer Manager |  3,000-5,000 POLY | 14 days starting the day of task acceptance

# Bounty Requirements

1. Module specs are fulfilled
2. Module is implemented using appropriate module interface
3. Module is tested with >95% branch coverage
4. Module is delivered before deadline

# Module Specs

1. Limit number of tokens which can be sent per transaction to a pre-defined limit. Revert if above limit.
2. Exempt special-use wallets from this volume restriction per trade.
3. Issuer should be able to set global limit for all addresses and set custom limit for each address.
4. Issuer should be able to change limit at any time.
5. Any state changes should emit an appropriate event.

# Assumptions
1. This volume restriction can operate in conjunction with all existing transfer restrictions and volume restrictions
2. Restrictions applies to all wallets by default
3. Some wallets (e.g. Treasury wallet, or Vesting wallet) may be subject to an exemption
4. Activation and/or Configuration of the "Single-Trade Volume Restriction" may be performed by an Authorized Employee of the Company or a Delegated Authority, such as a Transfer Agent
5. One or more Wallet addresses can be granted the privilege of managing the Single-Trade Volume Restriction.

# Detailed Module Specs
https://www.lucidchart.com/documents/view/ffbc99b9-a2d2-4043-8847-2bd0182aac3c 

**NOTE**: It would be ideal if the developer taking on this task could work on the single-trade vol restrictions from the development-1.5.0 branch instead of master.

subramanianv:

Closes #258 

### What is the new behavior?


pabloruiz55:
Thanks @subramanianv for starting to work on this. Let @adamdossa @satyamakgec or myself know if you have any questions. Seems like you are on the right direction so far ;)

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    if(paused) {
+      return Result.NA;
+    }
+
+    if(exemptWallets[_from]) return Result.NA;
+
+    if(specialTransferLimitWallets[_from]) {
+      if(globalTransferLimitInPercentage) {
+        validTransfer = (_amount.mul(100).div(ISecurityToken(securityToken).totalSupply())) <= specialTransferLimits[_from];
```
satyamakgec: Error in the calculation. You are giving the margin of 0.9 %. Suppose if the left size has the 15.7% and right side have the 15 % value then it still gets `true` instead of `false`.

subramanianv: @satyamakgec I don't fully understand your comment.  Is your comment because of this warning from the solidity docs ? ```Division on integer literals used to truncate in earlier versions, but it will now convert into a rational number, i.e. 5 / 2 is not equal to 2, but to 2.5.```

subramanianv: Also From PercentageTransferManager
`if(newBalance.mul(10**uint256(ISecurityToken(securityToken).decimals())).div(ISecurityToken(securityToken).totalSupply()) > maxHolderPercentage) {
                return Result.INVALID;
}`

Shouldn't newBalance be multipled by 10^18 explicitly instead of 10^decimals as maxHolderPercentage is multipled by (10^18/100) ?

satyamakgec: I don't think so because by default securitytoken have 18 decimals so in case of `PercentageTransferManager` it always give you the value right percentage calculation.
      
       newBalance = x * 10^d
       totalSupply = y * 10^d
       maxHolderPercentage = z * 10^d
       so it looks like this -  (x* 10^d * 10^d) / y* 10^d > (z * 10^d)/100      ... where d = 18 
       After calculation you can easily compare ((x * 10^18)/y) > (z* 10^16)

But in your case you are doing this 
     
       amount  = a* 10^d
       totalSupply = b* 10^d
       Limit= c* 10^d  (case 1)
       so it will look like this -  (a* 10^d * 100)/(b* 10^d) <= c* 10^d   .... where d = 18
       so your left hand side is always less than the right hand side.
       
       Limit = c (case 2)
       in this condition  - (155000 * 10^d * 100) / (1000000 * 10^d) <= 15   .... a= 155000, b = 1000000, c= 15 and d =18.
       After calculation it should be  15.5 <= 15  (and it will returns true as per my test on remix with 0.4.24 and 0.4.25 solidity version).

      and using case 2 you also can't cover the limits in decimal

subramanianv: @satyamakgec Thanks for the reply. The first case, which I haven't implemented is missing a multiplication by 10^18. it should be ```(a* 10^d * 10^18)/(b* 10^d) <= c* 10^16```. I will implement this formula/calculation.

In the `PercentageTransferManager` -  the math should be (x* 10^d * 10^18) / y* 10^d > (z * 10^18)/100. The multiplication should be explicitly done by 10^18. I think this would work for security token of any decimals, not just tokens with 18 decimals

satyamakgec:
@subramanianv I think the best time to review the code when you have done with the test cases and all the specs get covered.

pabloruiz55:
Hi @subramanianv thanks for submitting this PR. Could you please make it against the `development-1.5.0` branch instead? There might be some minimal changes to be made.

subramanianv:
@pabloruiz55 I made the PR against development-1.5.0. The PR is ready for review :)

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+  * @notice Used to intialize the variables of the contract
+  * @param _isTransferLimitInPercentage true if the transfer limit is in percentage else false
+  * @param _globalTransferLimitInPercentageOrToken transfer limit per single transaction.
+  */
+  function configure(bool _isTransferLimitInPercentage, uint256 _globalTransferLimitInPercentageOrToken) public onlyFactory {
+    require(_globalTransferLimitInPercentageOrToken > 0, "global transfer limit has to greater than 0");
+    isTransferLimitInPercentage = _isTransferLimitInPercentage;
```
pabloruiz55: if we are dealing with percentage require it to be <= 100

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    globalTransferLimit = _newGlobalTransferLimit;
+    emit LogGlobalTransferLimitInTokensSet(_newGlobalTransferLimit);
+  }
+
+  /**
+  * @notice Change the global transfer limit
+  * @param _newGlobalTransferLimitInPercentage new transfer limit in percentage. Multiple the percentage by 10^16. Eg 22% will be 22*10^16
+  */
+  function changeGlobalLimitInPercentage(uint256 _newGlobalTransferLimitInPercentage) public withPerm(ADMIN) {
+    require(isTransferLimitInPercentage, "Transfer limit not set in Percentage");
```
pabloruiz55: require 0 <= `_newGlobalTransferLimitInPercentage ` <= 100

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    specialTransferLimits[_wallet] = _transferLimit;
+    emit LogTransferLimitInTokensSet(_wallet, _transferLimit);
+  }
+
+  /**
+  * @notice set transfer limit per wallet an exempt wallet
+  * @param _wallet wallet address
+  * @param _transferLimitInPercentage transfer limit for the wallet in percentage. Multiple the percentage by 10^16. Eg 22% will be 22*10^16
+  */
+  function setTransferLimitInPercentage(address _wallet, uint _transferLimitInPercentage) public withPerm(ADMIN) {
```
pabloruiz55: Would be ideal to make this into a batch function. Pass _wallet and _transferLimitInPercentage as arrays and perform this operation for each item. So on the frontend we can perform these operations in batch.

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    specialTransferLimitWallets[_wallet] = true;
+    specialTransferLimits[_wallet] = _transferLimitInPercentage;
+    emit LogTransferLimitInPercentageSet(_wallet, _transferLimitInPercentage);
+  }
+
+  /**
+  * @notice removes transfer limit for a wallet
+  * @param _wallet wallet address
+  */
+  function removeTransferLimitForWallet(address _wallet) public withPerm(ADMIN) {
```
pabloruiz55: Same as above, please make into a batch function.

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    exemptWallets[_wallet] = false;
+    emit LogExemptWalletRemoved(_wallet);
+  }
+
+  /**
+  * @notice set transfer limit per wallet an exempt wallet
+  * @param _wallet wallet address
+  * @param _transferLimit transfer limit for the wallet in tokens
+  */
+  function setTransferLimitForWallet(address _wallet, uint _transferLimit) public withPerm(ADMIN) {
```
pabloruiz55: Would be ideal to make this into a batch function. Pass _wallet and _transferLimit as arrays and perform this operation for each item. So on the frontend we can perform these operations in batch.

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    require(isTransferLimitInPercentage, "Transfer limit not set in Percentage");
+    globalTransferLimitInPercentage = _newGlobalTransferLimitInPercentage;
+    emit LogGlobalTransferLimitInPercentageSet(_newGlobalTransferLimitInPercentage);
+  }
+
+  /**
+  * @notice add an exempt wallet
+  * @param _wallet exempt wallet address
+  */
+  function addExemptWallet(address _wallet) public withPerm(ADMIN) {
```
pabloruiz55: Same as other functions. Please make into batch function

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+    require(_wallet != address(0), "Wallet address cannot be a zero address");
+    exemptWallets[_wallet] = true;
+    emit LogExemptWalletAdded(_wallet);
+  }
+
+  /**
+  * @notice remove an exempt wallet
+  * @param _wallet exempt wallet address
+  */
+  function removeExemptWallet(address _wallet) public withPerm(ADMIN) {
```
pabloruiz55: Same as other function. Please make into batch function

contracts/modules/TransferManager/SingleTradeVolumeRestrictionManager.sol
```diff
@@ -0,0 +1,176 @@
...
+  // should be multipled by 10^16. if the transfer percentage is 20%, then globalTransferLimitInPercentage should be 20*10^16
+  uint256 public globalTransferLimitInPercentage;
+
+  mapping(address=>bool) public exemptWallets;
+
+  mapping(address => bool) public specialTransferLimitWallets;
+
+  mapping(address => uint) public specialTransferLimits;
+
+  event LogExemptWalletAdded(address _wallet);
```
pabloruiz55: Since some time ago solidity introduced the "emit" keyword, we recently changed all events so they don't start with "log", please remove "Log" prefix from every event.

pabloruiz55:
@subramanianv thanks for submitting the PR! In general terms it looks good! I like how you quickly got the hang out of how our modules have been structured. Left you a few requests for changes, basically making most of the functions so arrays are passed and multiple entries can be added/removed in a single tx. You can check other TM modules to see how we've done it.

On top of that, I'd need you to:
- Make sure natspec comments are fully added to each function
- Indentation follows Solidity Indentation style (Use 4 spaces per indentation level)

Thanks!

@satyamakgec could you check if there's anything else I missed?

subramanianv:
@pabloruiz55 I made all the requested changes

satyamakgec:
Thanks! @subramanianv for submitting the PR, could you please add more test cases related to the `SingleTradeVolumeRestrictionManager.sol` it has only `65.09 %` coverage. check the full result [here](https://coveralls.io/builds/19220012/source?filename=contracts%2Fmodules%2FTransferManager%2FSingleTradeVolumeRestrictionManager.sol#L135) Meanwhile I will take a second look on your code.

satyamakgec:
@subramanianv, I think it is better to give a function called `changeTransferLimitInPercentage()` to the issuer using this issuer can easily switch over the % restriction if an earlier state was on the `fix no. of tokens` or vice versa.  So please add this function.

Small suggestion that you can combine below two mappings
      
      mapping(address => bool) public specialTransferLimitWallets;
      mapping(address => uint) public specialTransferLimits;
                                        as
      mapping(address => WalletLimit ) public specialTransferLimits;
      struct WalletLimit {
         bool isAllowed;
         uint256 limit;
       }
or more easy way is use only one mapping `mapping(address => uint256) public specialTransferLimits;`
and `if (specialTransferLimits[address] == 0)` then it means address is not in the special category.

subramanianv:
@satyamakgec I added more tests now. I also removed the two mappings and now I am using one.

I added two functions `changeTransferLimitToPercentage` and `changeTransferLimitToTokens`. However with this, the user has to reconfigure the special wallets if they have any in the first place.

satyamakgec:
@subramanianv Thanks! Now the test coverage is very good. 
The only thing I am worried about the `specialTransferLimits` nature when Issuer switch between `%age` to  `fix no. of tokens` or vice versa. Don't you think it will be a pain to an issuer if he switches over then he needs to add the special limits again?  couldn't we maintain the different variables to hold the `%age` and `fix no. of token` that will make the life of issuer easier? @pabloruiz55 WDYT?

subramanianv:
@satyamakgec The cleanest approach would be to pause/remove the existing transfer manager and create a new transfer manager with the required config. That was my initial thought. We can also use two different variables to hold the specialtransferlimits as you suggested. lets wait for @pabloruiz55 to jump in.

satyamakgec:
Hey, @subramanianv you can wrap this up using the 2 variables approach.

subramanianv:
@satyamakgec I made into 2 variables. Please review the PR. Thanks :)

satyamakgec:
Hey @subramanianv  on logical side code is good to go. Please update your branch. 
Some minor fixes you need to add:
- [ ] Add the `comaptibleVersionRange` into your factory.
- [ ] Do the changes in your test cases as per the recent commit
- [ ] Add some comments/description for the state variables and event(ref. `SecurityToken.sol`).
- [ ] Change the test file name to `w_single_trade_volume_restriction.js` also remove all the external gas declaration from the test cases as we did in our latest commits.
- [ ] Add the module in migrations (we will see whether we are going to deploy it in our latest release or not. But for now you can add it).

subramanianv:
@satyamakgec I made all the requested changes.
> * [x]  Add the `comaptibleVersionRange` into your factory.
            Followed what is set in other contracts. `
        compatibleSTVersionRange["upperBound"] = VersionUtils.pack(uint8(0), uint8(0), uint8(0));`
> * [x]  Do the changes in your test cases as per the recent commit
             Done
> * [x]  Add some comments/description for the state variables and event(ref. `SecurityToken.sol`).
             Done
> * [x]  Change the test file name to `w_single_trade_volume_restriction.js` also remove all the external gas declaration from the test cases as we did in our latest commits.
      renamed to `w_single_trade_volume_restriction.js`. Removed all external gas
> * [x]  Add the module in migrations.
      Added the module in migrations

adamdossa (approved):

pabloruiz55:
@satyamakgec  there seems to be an issue with tests here, could you check?
```
Compilation failed. See above.
Cleaning up...
Event trace could not be read.
Error: ENOENT: no such file or directory, open './allFiredEvents'
Exiting without generating coverage...
```

satyamakgec (approved):

subramanianv:
@satyamakgec Can this PR be closed as the code has been merged into development ?

pabloruiz55:
Closing this PR as it was merged in #318 
Thanks @subramanianv !

PR #271, opened 2018-09-19
Add DAI to USDTieredSTO

adamdossa:
- Added DAI as an accepted currency for USDTieredSTO
  - Added daiToken as configuration param to USDTieredSTO
  - Updated tests including USDTieredSTO Sim
  - Updated CLI to reference new functions

### DIFF
--- a/CLI/commands/ST20Generator.js
+++ b/CLI/commands/ST20Generator.js
@@ -22,10 +22,14 @@ const regFee = 250;
 const cappedSTOFee = 20000;
 const usdTieredSTOFee = 100000;
 const tokenDetails = "";
+const ETH = 0;
+const POLY = 1;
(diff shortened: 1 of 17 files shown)


satyamakgec:
@adamdossa PR Looks good to me when all the test cases get passed then I will approve it.

adamdossa:
@VictorVicente I made updates to the CLI for the new functions, but I think accepting DAI for USDTieredSTO's will need more changes. We will need to decide what to do on ganache / testnet for DAI (probably easiest is to generate a mocked token, or re-use polyToken (current approach).

adamdossa:
@satyamakgec Test cases all run to success - please review.

satyamakgec (approved):

PR #276, opened 2018-09-24
WIP: New blacklist contract added

kpulkit29:
New blacklisting contract added Please explain what _perm means here in GeneralPermissionsManager.sol
function checkPermission(address _delegate, address _module, bytes32 _perm) external view returns(bool) {
        if (delegateDetails[_delegate] != bytes32(0)) {
            return perms[_module][_delegate][_perm];
        }else
            return false;
    }
Also module address here would be address of blacklisting contract right??
Please review the contract.Am I also required to write the subsequent test?


satyamakgec:
@kpulkit29  `checkPermission()` function is used to check whether the particular delegate address have the permission (_perm ex - ADMIN, WHITELIST or any perm flag that you created during your module creation) to  execute some function which has the `withPerm("WHITELIST")` modifier attached and yes `_module` is the address of the Module whose functions have `_perm` type access permission.

In crux `checkPermission() is use to check the permission on delegate corresponds to module contract address`

contracts/modules/PermissionManager/AutomatedBlackList.sol
```diff
@@ -0,0 +1,65 @@
+pragma solidity ^0.4.24;
+import "./GeneralPermissionManager.sol";
+import "../TransferManager/ITransferManager.sol";
+
+/**
+ * @title AutomateBlacklist
+ * @dev to automate blacklist process
+ */
+contract AutomateBlacklist is GeneralPermissionManager,ITransferManager {
```
satyamakgec: You don't need to inherit the `GeneralPermissionManager` contract. It will automatically available when an issuer will add this contract as a module using the `addModule()` in securityToken. To avail the functionality of permission issuer needs to add the GeneralPermissionManager module after or before adding the `AutomateBlacklistTransferManager` then only you can use the `changePermission()` function to add the perm to delegate corresponds to any module that is present in the module list of the securityToken.

contracts/modules/PermissionManager/AutomatedBlackList.sol
```diff
@@ -0,0 +1,65 @@
...
+        uint256 endDate;
+        uint256 repeats;
+    }
+ 
+     /** 
+     *@dev to check if msg.sender can add new entry
+     */
+    modifier isAllowed() {
+        require(msg.sender == securityToken || checkPermission(msg.sender,address(this),"1"));
+        _;
```
satyamakgec: Wrong usage of `checkPermission()` here use `withPerm(PERM)` modifier here present in the `Module.sol` contract.

contracts/modules/PermissionManager/AutomatedBlackList.sol
```diff
@@ -0,0 +1,65 @@
...
+    mapping(string => address[]) blacklistToAddress;
+    
+    /**
+     * @dev to check if transfer can be made
+     */
+    function verifyTransfer(address _from, address _to, uint256 _amount, bool _isTransfer) public returns(Result) {
+
+        string storage typeOfBlacklist=inverstorToBlacklist[_from];
+        require(bytes(inverstorToBlacklist[_from]).length!=0);
+        uint256 repetitions = (now / (blacklists[typeOfBlacklist].endDate - blacklists[typeOfBlacklist].startDate + blacklists[typeOfBlacklist].repeats));
```
satyamakgec: use `SafeMath` for the calculation

contracts/modules/PermissionManager/AutomatedBlackList.sol
```diff
@@ -0,0 +1,65 @@
...
+    }
+     /**
+      * @dev to add the new blacklist entry
+      * @param {string} _type
+      * @param {address} _investor
+      * @param {uint256} _start
+      * @param {uint256} _end
+      * @param {uint256} _repeat
+      */
+    function addEntry(string _type,address _investor,uint256 _start,uint256 _end,uint256 _repeat) isAllowed() public {
```
satyamakgec: I am not preferring this type of blacklist entry addition. 
Suppose there are 10 employees in a company and you are restricting all of them with the same date restrictions then it doesn't make sense to rewrite the state variables(`blacklists[_type]=BlacklistDetails(_start,_end,_repeat);`) again and again 

I prefer to add one function which only adds the blackList type extensively as I mentioned in the specs 
```function addBlackListType(uint256 _startDate, uint256 _endDate, bytes32 _name, uint256 _repeatPeriodInDays);```
and one function for the addition of the employees/affiliates corresponds to BlackList Type
```function addInvestorToBlackList(address _investor, bytes32 _blackListName);```

Have a look at the [lucidchart]( https://www.lucidchart.com/invitations/accept/c544a325-e74d-4e54-a360-71805a9f10ce) for exact requirements.

satyamakgec (changes requested):

satyamakgec:
@kpulkit29 

**Suggestions**: Please go through the existing modules first and understand the underlying architecture then go for the code. Take your time and we are not in a hurry you have 15 - 20 days to finish so I strongly suggest to go through the procedure first and if you need any help to understand the architecture then I am happy to help.

**Note:** Raised PR doesn't have the factory code to create the `AutomateBlacklistTransferManager.`

contracts/modules/PermissionManager/AutomatedBlackList.sol
kpulkit29: I think that the issuer should add GeneralPermissionManager first then the automated blacklisting one

satyamakgec: An issuer can add `GeneralPermissionManager` after adding the `AutomatedBlacklistTransferManager` but issuer can't use the permission layer until the GPM is added as a module.

PR #277, opened 2018-09-25
Withholding tax on dividends

adamdossa:
PR to add withholding tax is ready for review. I also pulled in minor updates from the master branch as part of this.


contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -53,53 +25,54 @@ contract EtherDividendCheckpoint is ICheckpoint, Module {
...
+        require(_expiry > _maturity, "Expiry is before maturity");
+        require(_expiry > now, "Expiry is in the past");
+        require(msg.value > 0, "No dividend sent");
         require(_checkpointId <= ISecurityToken(securityToken).currentCheckpointId());
         uint256 dividendIndex = dividends.length;
         uint256 currentSupply = ISecurityToken(securityToken).totalSupplyAt(_checkpointId);
+        uint256 excludedSupply = 0;
+        for (uint256 i = 0; i < _excluded.length; i++) {
+            excludedSupply = excludedSupply.add(ISecurityToken(securityToken).balanceOf(_excluded[i]));
+        }
```
satyamakgec, line 75: Shouldn't be here `excludedSupply.add(ISecurityToken(securityToken).balanceOfAt(_excluded[i], _checkpointId));` ?

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -164,15 +100,18 @@ contract EtherDividendCheckpoint is ICheckpoint, Module {
...
-        uint256 claim = calculateDividend(_dividendIndex, _payee);
+        (uint256 claim, uint256 withheld) = calculateDividend(_dividendIndex, _payee);
         _dividend.claimed[_payee] = true;
         _dividend.claimedAmount = claim.add(_dividend.claimedAmount);
-        if (claim > 0) {
-            if (_payee.send(claim)) {
-              emit EtherDividendClaimed(_payee, _dividendIndex, claim);
+        uint256 claimAfterWithheld = claim.sub(withheld);
+        if (claimAfterWithheld > 0) {
+            if (_payee.send(claimAfterWithheld)) {
```
satyamakgec, line 107: Is there any use case we are covering by using `send` instead of `transfer`? My concern is if the send will return false then dApp will not understand that it is a failure or successful EtherDividend claim because in both scenarios it gets the same event.

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
adamdossa: Fixed

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
adamdossa: The use-case we were trying to address is when sending multiple payments, not failing the whole batch if one of the payments failed.

The `send` is wrapped in an if statement which should manually revert corresponding changes (and emit a `EtherDividendClaimFailed` event) if the send fails.

There was a bug which I've just fixed as well and I added a test case for this as well.

@satyamakgec Can you think of any other security issues with this approach?

satyamakgec: understood the use case, I don't think so any more security issue with this approach as `send` and `transfer` works very similarly only difference is `send` return boolean and `transfer` revert so in case of send we need to reset all state variables if we get `false` (You already did that). So I think we are fine with this approach the only issue was the wrong event emission that you fixed.

satyamakgec (approved):

PR #285, opened 2018-09-27
Check pause flow

satyamakgec:

Merge only when the `upgrade-mr` PR get merge as I take that upgrade-mr as a base branch

### What is the new behavior?

### DIFF
--- a/contracts/ModuleRegistry.sol
+++ b/contracts/ModuleRegistry.sol
@@ -69,13 +69,21 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
     /**
      * @notice Modifier to make a function callable only when the contract is not paused.
      */
-    modifier whenNotPaused() {
+    modifier whenNotPausedOrOwner() {
         if (msg.sender == getAddress(Encoder.getKey("owner"))) 
           _;
         else {
             require(!getBool(Encoder.getKey("paused")), "Already paused");
             _;
         }
+    }
+
+    /**
+     * @notice Modifier to make a function callable only when the contract is not paused and ignore is msg.sender is owner.
+     */
+    modifier whenNotPaused() {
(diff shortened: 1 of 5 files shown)


pabloruiz55:
PR looks good. Only wondering why all those commits are showing up in the PR. @satyamakgec

contracts/SecurityTokenRegistry.sol
```diff
@@ -116,10 +116,15 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
      * @notice Modifier to make a function callable only when the contract is not paused.
      */
     modifier whenNotPaused() {
```
adamdossa: Can you rename to something like whenNotPausedOrOwner - `whenNotPaused` might be a bit confusing as it's a standard function with a different normal implementation.

contracts/ModuleRegistry.sol
```diff
@@ -296,7 +300,7 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
     /**
      * @notice Called by the owner to pause, triggers stopped state
      */
-    function pause() external whenNotPaused onlyOwner {
+    function pause() external onlyOwner {
```
adamdossa: I think `pause` should revert if the contract is already paused. It makes the logs easier to interpret. Can we have `whenNotPaused` and `whenNotPausedOrOwner` and use `whenNotPaused` here maybe?

adamdossa (changes requested):

satyamakgec:
@pabloruiz55 the cause of showing all other commits was `upgrade-mr` branch as I create this branch from the upgrade-mr branch. not it doesn't show those commits as we merge the upgrade-mr branch.

adamdossa (approved):

Issue #287, opened 2018-09-27
Module Development Task: Lockup Volume Restrictions


CPSTL:
# Module Development Task
# Lockup Volume Restrictions Module

Module Type | Bounty | Completion deadline
-- | -- | --
Transfer Manager | TBD | 14 days to complete once task is confirmed by external developer(s).

# Short description

A vesting schedule but the investor address needs to hold the tokens. As an Issuer, I want to be able to establish a Lockup period for my Affiliates and Employees so that they cannot materially impact the Security Token Price by selling all their Security Tokens in a short period of time. 

# Bounty Requirements
1. Module specs are fulfilled
2. Module is implemented using appropriate module interface
3. Module is tested with >95% branch coverage
4. Module is delivered before deadline
5. Module does not present any bugs.
6. Volume restriction until date
7. Volume restriction by time period
8. Volume restriction by transaction

# Volume restriction until date

Essentially this is a vesting schedule but the investor address needs to hold the tokens. 

This **can** be implemented as TM module with current architecture.

# Implementation

Vesting schedule for each investor is stored in the TM

``` 
require(amount <= balanceOf(from) - lockedAmount(from));
```
# Example

This applies to staff and current shareholders. For example, if the total number of tokens issued to staff John Doe is 500,000 these are locked-up and released over 4 years.

> Prior to 31/12/2019 John Doe cannot transfer any tokens
> After 31/12/2019 John Doe is allowed to transfer 25% or 125,000 tokens
> After 31/12/2020 John Doe is allowed to transfer 50% or 250,000 tokens
> After 31/12/2021 John Doe is allowed to transfer 75% or 375,000 tokens
> After 31/12/2022 John Doe is allowed to transfer 100% or 500,000 tokens

# Detailed Module Specs (Wireframe Chart)

**Wireframe Link:** https://www.lucidchart.com/invitations/accept/9a7e8e83-81c7-4f80-8e9e-c530fa8f2a63

satyamakgec:
resolved by PR #290

PR #290, opened 2018-09-28
Lockup Volume Restrictions

glitch003:
- [ X ] The commit message follows our Submission guidelines
- [ X ] Tests for the changes have been added (for bug fixes / features)
- [ ? ] Docs have been added / updated (for bug fixes / features)

https://github.com/PolymathNetwork/polymath-core/issues/287

No

### Any Other information:
Some notes:
- Users can have multiple lock ups.  The amount they can withdraw at any given time is the sum of the allowed amounts for all their lock ups.
- Evenly divisible parameters aren't enforced.  You can create a lockup with 2 year total vesting period and 1.5 year release frequency.  I can change this if desired but figured it might be enforced on a front end.  Once the total vesting period has passed, the lock up is not enforced at all anymore.
- Lock ups have a "start time".  Before this time begins, users can't transfer their tokens.  
- Where should I add docs?  

Thanks!


glitch003:
Heads up - this isn't ready yet.  I'm adding some parameter checking.

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+            }
+            return Result.INVALID;
+        }
+        return Result.NA;
+    }
+    
+    // lets the owner / admin create a volume restriction lockup.
+    // takes a userAddress and lock up params, and creates a lockup for the user.  See the LockUp struct def for info on what each parameter actually is
+    function addLockUp(address userAddress, uint lockUpPeriodSeconds, uint releaseFrequencySeconds, uint startTime, uint totalAmount) public withPerm(ADMIN) {
+
```
satyamakgec: I prefer to add some sanity checks for the input parameters

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+        // Get an array of the user's lockups
+        LockUp[] storage userLockUps = lockUps[userAddress];
+        
+        // create a new lock up and push it into the array of the user's lock ups
+        LockUp memory newLockUp;
+        newLockUp.lockUpPeriodSeconds = lockUpPeriodSeconds;
+        newLockUp.releaseFrequencySeconds = releaseFrequencySeconds;
+        newLockUp.startTime = startTime;
+        newLockUp.totalAmount = totalAmount;
+        userLockUps.push(newLockUp);
```
satyamakgec: Instead of the current approach, it can be re-write in a single line as 
`lockUps[userAddress].push(LockUP(lockUpPeriodSeconds, releaseFrequencySeconds, startTime , totalAmount))`

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+        uint startTime; // when this lockup starts (seconds)
+        uint totalAmount; // total amount of locked up tokens
+    }
+    
+    // maps user addresses to an array of lockups for that user
+    mapping (address => LockUp[]) internal lockUps;
+
+    enum LockUpOperationType { Add, Remove, Edit }
+
+    event LogModifyLockUp(
```
satyamakgec: As we change our event naming convention, So please remove the `Log` prefix from the name.
and I prefer the new event for operation type so it is easily understandable with the name. 
ex- 
`event  addNewLockUp(...)`
`event  removeLockUP(...)`
`event modifyLockUp(...)`

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+            toRemove.releaseFrequencySeconds,
+            toRemove.startTime,
+            toRemove.totalAmount,
+            LockUpOperationType.Remove
+        );
+
+        // move the last element in the array into the index that is desired to be removed.  
+        userLockUps[lockUpIndex] = userLockUps[userLockUps.length - 1];
+        // delete the last element
+        userLockUps.length--;
```
satyamakgec, line 158: Try to add a check 

         if (lockUpIndex < userLockUps.length - 1) {
             userLockUps[lockUpIndex] = userLockUps[userLockUps.length - 1];
           } 
          userLockUps.length--;

Using this we will save un-necessary state variable re-writing i.e save gas.

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+            addLockUp(userAddresses[i], lockUpPeriodsSeconds[i], releaseFrequenciesSeconds[i], startTimes[i], totalAmounts[i]);
+        }
+
+    }
+    
+    // remove a user's lock up
+    function removeLockUp(address userAddress, uint lockUpIndex) public withPerm(ADMIN) {
+        LockUp[] storage userLockUps = lockUps[userAddress];
+        LockUp memory toRemove = userLockUps[lockUpIndex];
+
```
satyamakgec, line 143: Add sanity check for the input parameters 
ex- 
require(lockUpIndex <= userLockUps.length - 1);

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+            LockUpOperationType.Remove
+        );
+
+        // move the last element in the array into the index that is desired to be removed.  
+        userLockUps[lockUpIndex] = userLockUps[userLockUps.length - 1];
+        // delete the last element
+        userLockUps.length--;
+    }
+
+    function editLockUp(address userAddress, uint lockUpIndex, uint lockUpPeriodSeconds, uint releaseFrequencySeconds, uint startTime, uint totalAmount) public withPerm(ADMIN) {
```
satyamakgec: rename the function name to `modifyLockUp()`

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+        
+        // if a startTime of 0 is passed in, then start now.
+        if (startTime == 0) {
+            startTime = now;
+        }
+
```
satyamakgec: sanity checks are missing

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+        }
+        
+        // Get the lockup from the master list and edit it
+        LockUp storage userLockUp = lockUps[userAddress][lockUpIndex];
+        
+        userLockUp.lockUpPeriodSeconds = lockUpPeriodSeconds;
+        userLockUp.releaseFrequencySeconds = releaseFrequencySeconds;
+        userLockUp.startTime = startTime;
+        userLockUp.totalAmount = totalAmount;
+
```
satyamakgec, line 188: can be done in a single line as I did in above comments

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
```diff
@@ -0,0 +1,241 @@
...
+                uint amountPerPeriod = aLockUp.totalAmount.div(totalPeriods);
+                // calculate the number of tokens that should be released,
+                // multiplied by the number of periods that have elapsed already
+                // and add it to the total allowedSum
+                allowedSum = allowedSum.add(amountPerPeriod.mul(elapsedPeriods));
+            }
+        }
+
+        // subtract already withdrawn amount
+        uint currentUserBalance = ISecurityToken(securityToken).balanceOf(userAddress);
```
satyamakgec, line 327: There is a high chance of mistake in calculation. there are many other possibilities where investor/Affiliate balance can change. I strongly suggest adding the mechanism which will keep track of withdrawn balance from the lockups.

satyamakgec (changes requested):

satyamakgec:
@glitch003 I miss reading the `Heads Up comment`. But I left some of the comments over the PR. Hope these will help you to add more checks and improve the code. 
**Note:** All functions missing `natspec comments`. So please add those as well.

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
glitch003: Hey @satyamakgec - could you clarify this one?  What kind of sanity check were you thinking?

It was my thinking that a startTime before "now" may be desired if you want to backdate someone's vesting, which I think is a common thing.  And obviously a startTime after "now" could be desired if you're awarding tokens to an employee who, for example, starts working in 2 weeks or something.

contracts/modules/TransferManager/VolumeRestrictionTransferManager.sol
glitch003: I understand, but, I'm not sure that the already withdrawn amount should be stored in this transfer module: 

1. verifyTransfer() is public.  it probably shouldn't mutate state internally, right?  
2. there's no way to know inside a transfer module if entire transfer succeeds, right?  another transfer manager might block the transfer.  Or, is there a success callback of some kind?

Maybe there is another place to store this already withdrawn amount?  On the security token itself?

adamdossa: Hi @glitch003,

The correct pattern here is to use the `_isTransfer` parameter of the function:  
`function verifyTransfer(address _from, address _to, uint256 _amount, bool _isTransfer) public returns(Result)`

If this flag is `true` it means that the transfer will actually happen (as opposed to someone checking whether a transfer is valid without actually executing the transfer). In other words, if `_isTransfer == false` the `verifyTransfer` function should act as a view function (i.e. not change state), but if `_isTransfer == true` the `verifyTransfer` function can maintain state related to this.

There is no way of expressing the above in the interface itself, but this is the intention for how these functions should behave.

If you are maintaining state, you'll need to add the below to the `verifyTransfer` function to deal with the fact that it is public:  
`require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");`

adamdossa: The `ManualApprovalTransferManager` does this if you want a concrete example to look at.

glitch003: @adamdossa okay, that definitely resolves my 1st point.  Thanks for explaining!

But I think my second point still stands, right?  What if a token has two transfer managers on it, and this Volume Restriction manager returns Result.VALID and marks a user as having withdrawn tokens.  But the next transfer manager returns Result.INVALID.  Therefore, the transfer never actually happens, but the Volume Restriction manager marked it as happening.  There's now a mismatch between the real world "amount withdrawn" and the amount stored in this Volume Restriction manager.

That's why I asked if there's a success callback of some kind that can tell the Volume Restriction module that the entire transfer succeeded, but I don't see one.  I don't think it would be that hard to add, though.

@satyamakgec or @CPSTL - any thoughts here?

adamdossa: If an actual transfer is happening (`_isTransfer == true`) and the transfer fails (i.e. someone returns INVALID) then the whole transaction is reverted, so any state updates you make would be reverted as well.

adamdossa: In other words you can assume that if the transfer doesn't actually happen, the whole transfer would be reverted.

glitch003: K, thanks.  Makes sense.  I made those changes.

glitch003:
@satyamakgec ok, this should be good to go.  I made the requested changes, and now the withdrawn amounts are tracked inside the transfer module.  

Let me know if there's anything else that needs to be changed

adamdossa (approved):

glitch003:
Hey @adamdossa thanks for those fixes to update my branch with dev-1.5.0.  Much appreciated!

I put in one more fix, and now the tests pass.

Let me know if there are any more changes needed to get this merged.  

Thanks!

satyamakgec (approved):

PR #291, opened 2018-09-28
Add better redemption mechanics

adamdossa:
Redemption Mechanics:

NB - redemption / burning is the same thing - currently we use burn rather than redeem as the verb.

Currently we have a burn function that references a ITokenBurner contract that can only be called by the user wishing to burn tokens.

Issues:
  - No ability to `approve` someone to burn tokens on your behalf
  - no way for issuer to burn tokens (forceTransfer doesn’t allow transfer to 0x0 or reducing the totalSupply)
  - no way for a module to burn tokens
  - not possible to have multiple burn mechanics

Requirements:
  - Token holders cannot directly burn tokens - must happen through a REDEMPTION type module
  - REDEMPTION type modules cannot arbitrarily burn tokens, users must approve the module first for the amount of tokens that they are ceding control over
  - Issuer can force burn tokens (reducing total supply) subject to the same controls as forceTransfer

Implementation:
  - Addition of `burnFrom` as an analogue of `transferFrom` - i.e. uses allowance to burn tokens on behalf of someone else
  - `burn` and `burnFrom` can only be called from a REDEMPTION type module
  - Addition of `forceBurn` to allow an issuer to burn someone elses tokens (subject to same controls as `forceTransfer`
  - Issuer cannot call `burn` or `burnFrom` (i.e. we use `onlyModule`, not `onlyModuleOrOwner`)
  - Remove `tokenBurner` as redemption logic will instead happen in REDEMPTION module
  - Renamed STO_KEY as MINT_KEY and added BURN_KEY as a module type

Notes:
  - One type of behaviour this doesn’t allow is forced burning via a REDEMPTION module. However the issuer can still force burn tokens.


contracts/tokens/SecurityToken.sol
```diff
@@ -541,25 +540,29 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
      * @param _value value of transfer
      * @return bool success
      */
-    function transferFrom(address _from, address _to, uint256 _value) public returns (bool success) {
+    function transferFrom(address _from, address _to, uint256 _value) public returns(bool) {
+        require(_updateTransfer(_from, _to, _value), "Transfer is not valid");
+        require(super.transferFrom(_from, _to, _value));
+        return true;
+    }
```
satyamakgec, line 544: we can call the `require(_updateTransfer(_from, _to, _value), "Transfer is not valid");` in transfer() function as well

contracts/tokens/SecurityToken.sol
```diff
@@ -616,85 +619,65 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-
-        for (uint8 i = 0; i < modules[PERMISSIONMANAGER_KEY].length; i++) {
-            if (IPermissionManager(modules[PERMISSIONMANAGER_KEY][i]).checkPermission(_delegate, _module, _perm)) {
-                return true;
-            }
-        }
+    function _burn(address _from, uint256 _value) internal returns (bool) {
+        require(_value <= balances[_from], "Value too high");
+        require(_updateTransfer(_from, address(0), _value), "Burn is not valid");
+        balances[_from] = balances[_from].sub(_value);
```
satyamakgec, line 654: missing function call `_adjustTotalSupplyCheckpoints();`

contracts/tokens/SecurityToken.sol
```diff
@@ -616,85 +619,65 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-
-        balances[msg.sender] = balances[msg.sender].sub(_value);
-        require(tokenBurner.burn(msg.sender, _value), "Token burner process is not validated");
-        totalSupply_ = totalSupply_.sub(_value);
-        emit Burnt(msg.sender, _value);
-        emit Transfer(msg.sender, address(0), _value);
+    function burnFrom(address _from, uint256 _value) checkGranularity(_value) onlyModule(BURN_KEY) public returns (bool) {
+        require(_value <= allowed[_from][msg.sender]);
+        require(_burn(_from, _value));
+        allowed[_from][msg.sender] = allowed[_from][msg.sender].sub(_value);
```
satyamakgec, line 677: I probably think we should deduct the allowed balance first and then call `_burn()` function. I am not 100 % sure but the code seems vulnerable to cross-function race conditions (`transferFrom()` and `burnFrom()`).

contracts/tokens/SecurityToken.sol
```diff
@@ -822,6 +822,23 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
+    /**
+     * @notice Use by a controller to execute a foced burn
+     * @param _from address from which to take tokens
+     * @param _value amount of tokens to transfer
+     * @param _data data attached to the transfer by controller to emit in event
+     */
+    function forceBurn(address _from, uint256 _value, bytes _data) public onlyController returns(bool) {
+        require(_value <= balances[_from], "Value too high");
+        bool verified = _updateTransfer(_from, address(0), _value);
+        balances[_from] = balances[_from].sub(_value);
```
satyamakgec, line 833: Here also need to call the`_adjustTotalSupplyCheckpoints()` ?

satyamakgec (changes requested):

satyamakgec (approved):
Looks good to me @pabloruiz55 please take a second look

contracts/libraries/KindMath.sol
```diff
@@ -0,0 +1,53 @@
+pragma solidity ^0.4.24;
+
+// Copied from OpenZeppelin and modified to be friendlier
+
+/**
+ * @title SafeMath
```
pabloruiz55: change title to KindMath

Issue #268, opened 2018-09-19
Module Development Task: Volume Restriction (Dumping)


CPSTL:
# Module Development Task
# Volume Restriction (Dumping)

Module Type | Bounty | Completion deadline
-- | -- | --
Transfer Manager | Approx. 10, 000 POLY | 21 days to complete once task is confirmed by external developer(s).

# Short Summary: 
This requires a way to track the number of tokens sent in a given time period. This module aims to prevent the dumping ability for security token owners by restricting trading by time period. 

**Note:** This can be implemented as TM module with current architecture. 

# Bounty Requirements
1. Module specs are fulfilled
2. Module is implemented using appropriate module interface
3. Module is tested with >95% branch coverage
4. Module is delivered before deadline
5. Module does not present any bugs.
6. 21 days to complete module from developers initial start date.

**IMPORTANT:** Please work from the development-1.5.0 branch instead of master.

# Module Specs and Suggested Implementation
The implementation requires keeping count of volume sent from an address in a given time period.

```
mapping(address => mapping(uint256 => uint256)) volumeTally;
mapping(address => mapping(uint256 => uint256)) volumeLimit;
mapping(address => uint256) limitPeriod;

// Assumes this function is only called when the transfer is successful
function verifyTransfer(address _from, address _to, uint256 _amount, bool _isTransfer) public returns(Result) {
    uint256 periodId = now.div(limitPeriod[_from]);
    if (_amount.add(volumeTally[_from][periodId]) > volumeLimit[_from][periodId]) {
        return false;
    };
    volumeTally[_from][periodId] += _amount;
    return true;
}
```

# Module Details and Specs (Wireframe) 

**Wireframe Link:** https://www.lucidchart.com/invitations/accept/8fcf7c3f-6f84-4846-8e5c-14a7df29a8cc

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 10000.0 POLY (1499.0 USD @ $0.15/POLY)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/PolymathNetwork/polymath-core/268/1257).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/PolymathNetwork/polymath-core/268/1257).
[...]

PR #292, opened 2018-09-28
Volume dumping restriction manager

samparsky:
- [x] Docs have been added / updated (for bug fixes / features)

#268 

### What is the new behavior?

No


satyamakgec:
Hey, @samparsky Should I start reviewing this PR or still you are adding some code in it?

samparsky:
@satyamakgec Currently adding more tests. Will let you know when I am done

samparsky:
@satyamakgec Please kindly review this PR

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,321 @@
+pragma solidity ^0.4.24;
+
+import "./ITransferManager.sol";
+import "../../tokens/SecurityToken.sol";
```
satyamakgec: No requirement of importing the SecurityToken.sol here

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,321 @@
...
+        uint256 startTime,
+        uint256 endTime,
+        uint256 rollingPeriod
+
+    );
+
+    event Testing (
+        address indexed from, 
+        uint256 text
+    );
```
satyamakgec: Unnecessary code. - remove it

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,321 @@
...
+    Module(_securityToken, _polyAddress)
+    {
+    }
+
+    /** @notice Used to verify the transfer transaction and prevent locked up tokens from being transferred
+     * @param _from Address of the sender
+     * @param _amount The amount of tokens to transfer
+     */
+    function verifyTransfer(address _from, address /*_to */, uint256 _amount, bool /* _isTransfer */) public returns(Result) {        
+        if (!paused && _from != address(0)) {
```
satyamakgec: Below code, you are changing the state so it means anybody can change it because `verifyTransfer()` is public function. To solve this you should apply a restriction as `require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");`. for concrete example you can take a look into the `ManualApproavalTransferManager.sol`

satyamakgec (changes requested):

satyamakgec:
@samparsky left some comments on the PR. and you can also improve the test coverage of your contracts although it is good please try to make it `95% - 98%`

samparsky:
@satyamakgec I have made the changes suggested above.
The test seems to be failing due to some issue I can't figure out

This returns 
```sh
$ node_modules/.bin/solidity-coverage
```
the following error which seems odd, given that `truffle compile` works perfectly
```
* Line 1, Column 1
  Syntax error: value, object or array expected.
* Line 1, Column 1
  A valid JSON document must be either an array or an object value.
```

satyamakgec:
Yes @samparsky we are facing this issue, Once we resolve that then your PR also start building in Travis.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,322 @@
...
+                return Result.VALID;
+            }
+            
+            uint256 periodId = now.div(userDumpingRestriction.rollingPeriod);
+
+            // // the remainig transferable amount
+            uint256 allowedRemainingAmount = _currentAllowedAmount(_from, periodId);
+
+            if (_amount <= allowedRemainingAmount){
+                volumeTally[_from][periodId] = volumeTally[_from][periodId].add(_amount);
```
satyamakgec: Here you are changing the state variable. It should be restricted and only be performed when `(_isTransfer == true or securityToken == msg.sender)`. You are allowing the change when anyone going to query the public `verifyTransfer()` function

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,322 @@
...
+        require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");
+
+        if (!paused && _from != address(0)) {                
+            VolumeRestriction memory userDumpingRestriction = volumeRestriction[_from];
+
+            if (now < userDumpingRestriction.startTime || now > userDumpingRestriction.endTime){
+                return Result.VALID;
+            }
+            
+            uint256 periodId = now.div(userDumpingRestriction.rollingPeriod);
```
satyamakgec: I don't think it is a right approach. For correct periodID, it should be 
`uint256 periodId = (now - startTime).div(userDumpingRestriction.rollingPeriod)`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,322 @@
...
+     /**
+    * @notice lets the admin remove a volume dumping restriction.
+    * @param _userAddress User address to remove the volume dumping restriction
+    */
+    function removeRestriction(address _userAddress) public withPerm(ADMIN) {
+        VolumeRestriction memory toRemove = volumeRestriction[_userAddress];
+
+        // check if entry is in the past
+        // not allowed to delete past entries
+        require(toRemove.endTime > now, "Cannot remove past restrictions");
```
satyamakgec: wrong reason string. It should be `Only past restrictions are allowed to delete` more preferable is `Only passed restrictions are allowed to delete`

satyamakgec (changes requested):

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: According to the flow chart design in the [Lucid Chart](https://www.lucidchart.com/documents/edit/c96ff64c-5b89-4b9c-985a-01637266bf88/0?shared=true&) provided It says below
image

It would contradict it its essentially checking if the restriction being removed endtime is still current. Using `Only past restrictions are allowed to delete` would require changing it to 

```javascript   
require(toRemove.endTime < now, "Only past restrictions are allowed to delete");
```

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: Are there scenarios when `isTransfer` would be `false`? like a readonly txn?

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
satyamakgec: Hmm agreed keep as it is

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
satyamakgec: Yes for readonly `isTransfer` should be false.

samparsky:
@satyamakgec Please review, the changes suggested have been implemented

samparsky:
@satyamakgec Please review, the changes suggested have been implemented

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+
+        if(!_isTransfer) {
+            return Result.VALID;
+        }
+
+        if (!paused && _from != address(0)) {                
+            VolumeRestriction memory userDumpingRestriction = volumeRestriction[_from];
+
+            if (now < userDumpingRestriction.startTime || now > userDumpingRestriction.endTime){
+                return Result.VALID;
```
adamdossa: This should return Result.NA not Result.VALID (if the `_from` address doesn't have any dumping restrictions it doesn't mean the transfer is necessarily valid, just that it is not invalid from the perspective of this transfer manager)

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+            }
+            
+            uint256 periodId =  now.sub(userDumpingRestriction.startTime).div(userDumpingRestriction.rollingPeriod);
+
+            // // the remainig transferable amount
+            uint256 allowedRemainingAmount = _currentAllowedAmount(_from, periodId);
+
+            if (_amount <= allowedRemainingAmount){
+                volumeTally[_from][periodId] = volumeTally[_from][periodId].add(_amount);
+                return Result.VALID;
```
adamdossa: As above should return Result.NA here not Result.VALID

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+
+     /**
+    * @notice Takes a userAddress as input, and returns a uint that represents the number of tokens allowed to be withdrawn right now
+    * @param _userAddress User Address to get the rule
+    * @param periodId Time period identifier
+    */
+    function _currentAllowedAmount(address _userAddress, uint periodId) internal returns (uint) {
+        VolumeRestriction storage userDumpingRestriction = volumeRestriction[_userAddress];
+        
+        uint alreadyWithdrawn = volumeTally[_userAddress][periodId];
```
adamdossa: Can we consistently use `uint256`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+        allPermissions[0] = VOLUME_RESTRICTION;
+        return allPermissions;
+    }
+
+     /**
+    * @notice Takes a userAddress as input, and returns a uint that represents the number of tokens allowed to be withdrawn right now
+    * @param _userAddress User Address to get the rule
+    * @param periodId Time period identifier
+    */
+    function _currentAllowedAmount(address _userAddress, uint periodId) internal returns (uint) {
```
adamdossa: Should be `_periodId`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+    function verifyTransfer(address _from, address /*_to */, uint256 _amount, bytes /* _data */, bool _isTransfer) public returns(Result) {
+        
+        // function must only be called by the associated security token if _isTransfer == true
+        require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");
+
+        if(!_isTransfer) {
+            return Result.VALID;
+        }
+
+        if (!paused && _from != address(0)) {
```
adamdossa: Probably should also check that `_amount != 0` otherwise you'll add duplicate periods in `_currentAllowedAmount` I think.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,326 @@
...
+        // means its essentially a new period
+        if(alreadyWithdrawn == 0){
+            volumePeriodIds[_userAddress].push(periodId);
+        }
+
+        uint256 currentUserBalance = ISecurityToken(securityToken).balanceOf(_userAddress);
+        // add already withdrawn amount for this period 
+        // to be able to get the correct value
+
+        uint256 percentAllowed = userDumpingRestriction.percentAllowed;
```
adamdossa: For better accuracy percentages should be represented as [percentage] * 10**16. So 100% == 10**18, 50% = 0.5 * 10**18

adamdossa (changes requested):

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: When should `Result.VALID` be used?

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: It doesn't add duplicate periodId, since I am getting the `remaining amount` allowed to be withdrawn. But I have added the check though.

samparsky:
@adamdossa  @satyamakgec Please review, the changes suggested have been implemented

satyamakgec:
@samparsky Thanks for all your hard work and patience. we are in the process of reviewing the code. it may take a couple of days to finalize the module. So please bear me. I will try to send you an update as soon as possible.

samparsky:
@satyamakgec @adamdossa  It would be helpful if this PR can be reviewed & merged.
Merge conflicts occurs when other PRs get merged in. I have had to rebase & fix several times cos of this, it gets frustrating after some time.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+     * @param _from Address of the sender
+     * @param _amount The amount of tokens to transfer
+     */
+    function verifyTransfer(address _from, address /*_to */, uint256 _amount, bytes /* _data */, bool _isTransfer) public returns(Result) {
+        
+        // function must only be called by the associated security token if _isTransfer == true
+        require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");
+
+        if(!_isTransfer) {
+            return Result.VALID;
```
satyamakgec: This means it always returns valid when the `_isTrasfer` is false.  It doesn't sound well because there are higher chances to fail when the user does the actual transaction.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+    mapping(address => VolumeRestriction) internal volumeRestriction;
+
+    // Maps user address to a periodId & amount transferred so far in that periodId
+    mapping(address => mapping(uint256 => uint256)) internal volumeTally;
+
+    // map user address to a list of periodIds
+    mapping(address => uint256[]) internal volumePeriodIds;
+    
+    event ModifyVolumeDumping (
+        address indexed from,
```
satyamakgec: rename the `from` variable because it sounds confusing as it looks the address which modifies the dumping restriction. So make it `restrictedAddress` or `tokenHolder`. Do the same in all other events.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+        for (uint256 i = 0; i < _userAddresses.length; i++) {
+            addDumpingRestriction(_userAddresses[i], _percentsAllowed[i], _startTimes[i], _endTimes[i], _rollingPeriods[i]);
+        }
+    }
+
+     /**
+    * @notice lets the admin remove a volume dumping restriction.
+    * @param _userAddress User address to remove the volume dumping restriction
+    */
+    function removeRestriction(address _userAddress) public withPerm(ADMIN) {
```
satyamakgec: Add batch version of it

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+        address _userAddress, 
+        uint256 _percentAllowed, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriod
+        ) internal view {
+        require(address(0) != _userAddress, "Invalid user address");
+        require(_percentAllowed > 0 && _percentAllowed <= 100, "Invalid percent input");
+        require(_startTime == 0 || _startTime >= now, "Invalid start time");
+        require(_endTime > 0 && _endTime > _startTime, "Invalid endtime");
```
satyamakgec: no need of `_endTime > 0` when you are checking `_endTime> _startTime` and `require(_startTime == 0 || _startTime >= now, "Invalid start time")`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+
+        _checkVolumeDumpingParams(_userAddress, _percentAllowed, _startTime, _endTime, _rollingPeriod);
+ 
+        require(volumeRestriction[_userAddress].endTime > now, "Cannot modify past restrictions");
+
+        if (_startTime == 0) {
+            _startTime = now;
+        }
+
+        require(_endTime > _startTime, "End time must be greater than start time");
```
satyamakgec: no need of this when you already checking all values in `_checkVolumeDumpingParams()`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+        uint256 _rollingPeriod
+        ) public withPerm(ADMIN) {
+
+        _checkVolumeDumpingParams(_userAddress, _percentAllowed, _startTime, _endTime, _rollingPeriod);
+ 
+        require(volumeRestriction[_userAddress].endTime > now, "Cannot modify past restrictions");
+
+        if (_startTime == 0) {
+            _startTime = now;
+        }
```
satyamakgec: move this if statement at the top of the function before the `_checkVolumeDumpingParams()`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+    
+    /**
+    * @notice lets the admin modify volume dumping restriction for an addresses.
+    * @param _userAddress Address of the user to apply the volume dumping restriction
+    * @param _percentAllowed Percent of tokens balance allowed to transfer within a rolling period
+    * @param _startTime When the dumping restriction, 0 means now in seconds
+    * @param _endTime When the dumping restriction ends in seconds
+    * @param _rollingPeriod is the time period in seconds
+    */
+    function modifyVolumeDumpingRestriction(
```
satyamakgec: Add a batch version of this function i.e - `modifyVolDumpingRestricionMulti()`

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+        // means its essentially a new period
+        if(alreadyWithdrawn == 0){
+            volumePeriodIds[_userAddress].push(_periodId);
+        }
+
+        uint256 currentUserBalance = ISecurityToken(securityToken).balanceOf(_userAddress);
+        // add already withdrawn amount for this period 
+        // to be able to get the correct value
+
+        uint256 percentAllowed = userDumpingRestriction.percentAllowed.mul(uint256(10)**16);
```
satyamakgec: This multiplication should not be here. it should check at the addition time otherwise issuer not be able to put the percentage in decimals ex- 12.5 %, 35.25% etc.

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
...
+
+        // totalAmountAllowedForPeriod = currentbalance * 20/100
+        uint256 totalAmountAllowedForPeriod = currentUserBalance.add(alreadyWithdrawn).mul(percentAllowed).div(100 * uint256(10)**16);
```
satyamakgec: This will give you a wrong calculation 
ex- A will have the balance of 1000 tokens 
      percentage allowed - 10%  = 100 tokens 
      First txn deduct the 50 tokens and after first txn A will get the 1000 tokens from B.
 for the second transaction(same periodId) calculation will look like this 
    (1950 + 50 ) * (10)/100 -- it is totally wrong because now it will allow him to deduct 200 tokens instead of more 50 tokens.

contracts/modules/TransferManager/VolumeDumpingRestrictionManagerFactory.sol
```diff
@@ -0,0 +1,102 @@
...
+     * @notice Constructor
+     * @param _polyAddress Address of the polytoken
+     */
+    constructor (address _polyAddress, uint256 _setupCost, uint256 _usageCost, uint256 _subscriptionCost) public
+      ModuleFactory(_polyAddress, _setupCost, _usageCost, _subscriptionCost)
+    {
+        version = "1.0.0";
+        name = "VolumeDumpingTransferManager";
+        title = "Volume Dumping Transfer Manager";
+        description = "Manage the volume of tokens dumpable by an owner";
```
satyamakgec: add the compatible protocol version range. Please check the other module factories you can easily understand what I am trying to say

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
```diff
@@ -0,0 +1,325 @@
+pragma solidity ^0.4.24;
+
+import "./ITransferManager.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+/**
+ * @title  Transfer Manager module for restricting volume dumping within a time period for an account
+ */
+
+contract VolumeDumpingRestrictionManager is ITransferManager {
```
satyamakgec: change the contract name to `VolumeDumpingRstrictionTM`

migrations/2_deploy_contracts.js
```diff
@@ -9,6 +9,7 @@ const ERC20DividendCheckpointFactory = artifacts.require('./ERC20DividendCheckpo
 const ModuleRegistry = artifacts.require('./ModuleRegistry.sol');
 const ModuleRegistryProxy = artifacts.require('./ModuleRegistryProxy.sol');
 const ManualApprovalTransferManagerFactory = artifacts.require('./ManualApprovalTransferManagerFactory.sol')
+const VolumeDumpingRestrictionTransferManagerFactory = artifacts.require('./VolumeDumpingRestrictionManagerFactory.sol')
```
satyamakgec: Please remove from the migrations

migrations/2_deploy_contracts.js
```diff
@@ -161,6 +162,10 @@ module.exports = function (deployer, network, accounts) {
     // D) Deploy the PercentageTransferManagerFactory Contract (Factory used to generate the PercentageTransferManager contract use
     // to track the percentage of investment the investors could do for a particular security token)
     return deployer.deploy(PercentageTransferManagerFactory, PolyToken, 0, 0, 0, {from: PolymathAccount});
+  }).then(() => {
+    // D) Deploy the VolumeDumpingRestrictionTransferManagerFactory Contract (Factory used to generate the VolumeDumpingRestrictionTransferManager contract use
+    // to track the dumping crietria an investor could do for a particular security token)
+    return deployer.deploy(VolumeDumpingRestrictionTransferManagerFactory, PolyToken, 0, 0, 0, {from: PolymathAccount});
```
satyamakgec: Please remove from the migrations

scripts/test.sh
```diff
@@ -63,7 +63,7 @@ start_testrpc() {
   if ! [ -z "${TRAVIS_PULL_REQUEST+x}" ] && [ "$TRAVIS_PULL_REQUEST" != false ]; then
     node_modules/.bin/testrpc-sc --gasLimit 0xfffffffffff --port "$testrpc_port" "${accounts[@]}" > /dev/null &
   else
-    node_modules/.bin/ganache-cli --gasLimit 8000000 "${accounts[@]}" > /dev/null &
+    node_modules/.bin/ganache-cli --gasLimit 0xfffffffffff "${accounts[@]}" > /dev/null &
```
satyamakgec: We want to run all the txn within the Mainnet gas range i.e 8 MN

satyamakgec (changes requested):
Also, try to increase the branch coverage it is 80 % but we are expecting b/w 95% - 100%

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: This would be violating the design document shared with me. According to https://www.lucidchart.com/documents/edit/c96ff64c-5b89-4b9c-985a-01637266bf88/0?shared=true&

image

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: Does this mean the percentage values passed in `addRestriction`would be in the format e.g.
12 % = 12 * 10 ** 16
12.5 = 125 * 10 ** 15

satyamakgec: Yes, the value will always between 1 * 10^16 - 100 * 10^16.
12.5 = 125 * 10^15

contracts/modules/TransferManager/VolumeDumpingRestrictionManager.sol
samparsky: @satyamakgec Can you help confirm this?

samparsky:
@satyamakgec I have implemented the changes you have suggested. 
The code coverage for the files added is at 96% & 98%.

VolumeDumpingRestrictionTMFactory.sol - 96%
VolumeDumpingRestrictionTM.sol - 98%


I still have some questions,
Regarding this below, I haven't gotten any clear response from you yet as it doesn't quite agree with the design I was given https://www.lucidchart.com/documents/edit/c96ff64c-5b89-4b9c-985a-01637266bf88/0?shared=true&

```
This will give you a wrong calculation
ex- A will have the balance of 1000 tokens
percentage allowed - 10% = 100 tokens
First txn deduct the 50 tokens and after first txn A will get the 1000 tokens from B.
for the second transaction(same periodId) calculation will look like this
(1950 + 50 ) * (10)/100 -- it is totally wrong because now it will allow him to deduct 200 tokens instead of more 50 tokens.
```
It would be awesome to get a response

satyamakgec:
@samparsky Thanks for the changes Yes you are right it is not matching with the design. I will give you more information after the internal discussion on design.

samparsky:
@satyamakgec @adamdossa  I am waiting on your decision

satyamakgec:
Hey @samparsky sorry for the delay. The logic looks as - 
let’s assume the limit is 10,000 tokens in a rolling 30 days. If SUM (all sales in the last 29 days) + sale requested now < 10,000 then the transaction can proceed, else the transaction reverts. and yes you are doing right as per the lucidchart. Always takes the user current balance in the calculation. But please keep in mind the logic of rolling period which seems off to me in your current code.

samparsky:
@satyamakgec That means the calculation of the limit  is correct,

As for the logic of the rolling period is calculated thus

The assumption is there is no rollover of tokens transfer i.e. if you have 10k limit in a period & you only transfer 5k you don't get the remaining 5k rolled over to the next period

endtime - 200
starttime - 100
rollingperiod - 10

currentTime = 101
unit256 periodId = (101-100) / 10 = 0

currentTime = 102
uint256 periodId = (102 -100) / 10 = 0

currentTime = 120
uint256 periodId = (120-100) / 10 = 2

currentTime = 150
uint256 periodId = (150-100) / 10 = 5

It would be great if we can work together to close the PR asap, its taken quite a very long time.

samparsky:
@satyamakgec @adamdossa Still waiting on you

satyamakgec:
Hey we are at Devcon so the coming week will be pretty much busy. I will
catch you up after the devcon.
Cheers

samparsky:
@satyamakgec Hello, how was Devcon? Hoping we can conclude on this, this week

satyamakgec:
@samparsky Devcon was good and yes we can conclude this week.

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
```diff
@@ -0,0 +1,366 @@
...
+    /**
+    * @notice lets the admin remove a volume dumping restriction.
+    * @param _userAddress User address to remove the volume dumping restriction
+    */
+    function removeRestriction(address _userAddress) public withPerm(ADMIN) {
+        VolumeRestriction memory toRemove = volumeRestriction[_userAddress];
+
+        // check if entry is in the past
+        // not allowed to delete past entries
+        require(toRemove.endTime > now, "Cannot remove past restrictions");
```
satyamakgec, line 192: It doesn't make sense to delete the past entries. If issuer accidentally add someone then no way to remove the restriction.

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
```diff
@@ -0,0 +1,366 @@
...
+        uint256 _rollingPeriod
+        ) public withPerm(ADMIN) {
+
+        if (_startTime == 0) {
+            _startTime = now;
+        }
+
+        _checkVolumeDumpingParams(_userAddress, _percentAllowed, _startTime, _endTime, _rollingPeriod);
+ 
+        require(volumeRestriction[_userAddress].endTime > now, "Cannot modify past restrictions");
```
satyamakgec, line 291: same if the issuer wants to modify the details for a particular investor. He can't do it until the dumping period gets completed. But please do care with your logic because if Issuer changes details during the dumping period then it will affect the calculation of rolling period. 
Two ways to mitigate - 
1. Allow the ADMIN to modify only when startTime > now.
2. Change the logic of periodID to make it more dynamic.

satyamakgec (changes requested):

satyamakgec:
@samparsky please do resolve the above issues and update your branch so I can give this module a green light

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
samparsky: Yeah, that's what is implemented.

If the `endTime > now`, this means 
i.e. endTime = 150 & currentTime = 100.
This means that the restriction is still current so you can remove it but once it's not
i.e. endTime = 150 & currentTime = 200
Then the restriction is in the past and you can't modify it

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
samparsky: Okay then, I would add a restriction to allow to modify only when `startTime < now`

samparsky: I included that if the startTime hasn't been reached then you can modify the start time else you can't.

Can we clarify using the below example

For example if a user has the following config 
percentAllowed = 10
rollingPeriod = 30

And then intends to change it to
percentAllowed = 20
rollingPeriod = 30

It doesn't really affect anything

But if he changes it to
percentAllowed = 10
rollingPeriod = 40

If the rolling period has started then definitely it would affect the calculation, since the rolling period & start time is used in calculating the period id, 

Hence an approach would be to check the already spent amount in the current active auction then roll over the spent amount to the new periodId calculated.

I also would have to delete all previously existing periodIds to prevent a case of periodId collision

That would seem to take care of the edge case

Or is there anything else I am missing?

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
samparsky: `If issuer accidentally add someone then no way to remove the restriction.` As far has the endTime hasn't been exceeded the admin can remove the restriction.

satyamakgec: Sorry My bad, I misread

contracts/modules/TransferManager/VolumeDumpingRestrictionTM.sol
satyamakgec: Yes, you are covering all the base but I am not a fan of deleting the previous data because if any reason Issuer wants to get the details from the contract then Issuer receives fewer data.

samparsky: Yeah another approach would be to delete the only the periodIds greater than the newPeriodID, to prevent the id collision

samparsky: Or we can prevent changing the rollingPeriod & startTime since that's what is used in calculating the periodId once the restriction period has started
So if any need arises to modify the rolling period the issuer & the user would have to use a new address.

samparsky:
@satyamakgec I have implemented the suggested changes.  I changed the implementation of the `periodId` to be more dynamic to allow changing the rollingPeriod without affecting the periodId calculation.

maxsam4:
Bounty paid. Thanks a lot for your contribution @samparsky  .

samparsky:
@maxsam4 Any reason why the PR wasn't merged?

maxsam4:
@samparsky We have decided to merge this module and https://github.com/PolymathNetwork/polymath-core/pull/320

It would not have been fair of us to ask you guys to do the required change so we'll be doing them ourselves. 
If you want to work on merging them in your free time, feel free and I'll reopen this PR. Thank you for all your effort, we really appreciate your work.

PR #294, opened 2018-09-30
Optimise STR

adamdossa:
Optimise STR code size & fix modifyTicker bug


adamdossa:
@bakii

Please could you carefully review this PR as it fixes some bugs and makes some changes.

Previous bug was in `modifyTicker` - if you passed in a new ticker reference, it wouldn’t record this in `userToTickers` due to the logic in `_modifyTicker`

This was missed in the test cases, specifically the test case “Should add the new custom ticker” did not correctly record ownership. When we then transferred ownership in “Should change the details of the existing ticker” we actually then transfer the ownership of the wrong token since “ETH” does not have an index in “tickerIndex”.

We didn’t catch this in the getter test case “Should get the tickers by owner” as it checked that `account_temp` should have 2 tokens instead of 3. For this test case we should have:

`token_owner`: [DET, AAA, ETH, POLY]
`account_temp`: [DET2, LOG, LOG2]

adamdossa:
@satyamakgec  Also fixed a few cases where we weren't using the correct upper case ticker value - please could you double check these as well.

satyamakgec:
Ah! I miss it, actually the cause of bug was that I focused on solving the `modifySecurityToken()` case in which we are calling `_modifyTicker()` but those actions also affect the real functionality of `modifyTicker()`. Thanks for spotting this bug.

contracts/SecurityTokenRegistry.sol
```diff
@@ -626,22 +625,16 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
...
-    function _setProtocolVersion(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) internal {
-        uint8[] memory _version = new uint8[](3);
-        _version[0] = _major;
-        _version[1] = _minor;
-        _version[2] = _patch;
-        uint24 _packedVersion = VersionUtils.pack(_major, _minor, _patch);
-        require(VersionUtils.isValidVersion(getProtocolVersion(), _version),"In-valid version");
-        set(Encoder.getKey("latestVersion"), uint256(_packedVersion));
+    function setProtocolVersion(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) public {
+        uint8[] memory version = new uint8[](3);
```
satyamakgec: I don't know why you make it public? It should only be called by the owner otherwise anybody will change the ST contract logic.
So we need both types of function `internal`(called from the `initialize()`) and the `onlyOwner`.

contracts/SecurityTokenRegistry.sol
```diff
@@ -227,20 +228,18 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
      * @notice Internal -- Modifies the ticker details.
      */
     function _modifyTicker(address _owner, string _ticker, string _tokenName, uint256 _registrationDate, uint256 _expiryDate, bool _status) internal {
-        address currentOwner = getAddress(Encoder.getKey("registeredTickers_owner", _ticker));
-        if (currentOwner == address(0) && _registrationDate == 0 && _expiryDate == 0) {
-            _addTicker(_owner, _ticker, _tokenName, now, now.add(getUint(Encoder.getKey("expiryLimit"))), _status, true);
-            return;
+        address currentOwner = _tickerOwner(_ticker);
+        if (currentOwner != address(0)) {
```
satyamakgec, line 235: This logic will consume more gas when we are only changing some details of the ticker let's say tokenName, dates etc but not the owner.
Then according to this logic, it will first delete the ownership from the same owner under this logic 
       
              if (currentOwner != address(0)) {
                _deleteTickerOwnership(currentOwner, _ticker);
               }
 and then assign the ownership to the same owner using this 
     
              _addTicker(_owner, _ticker, _tokenName, _registrationDate, _expiryDate, _status, true);

I don't know how many times we will use that function but It is worth to add the logic that will consume less gas.

satyamakgec: Another condition is missing in the `_modifyTicker()` related to `modifySecurityToken()` function

          // If ticker didn't previously exist, it will be created
        _modifyTicker(_owner, ticker, _name, getUint(Encoder.getKey("registeredTickers_registrationDate", ticker)), getUint(Encoder.getKey("registeredTickers_expiryDate", ticker)), true);

According to the current logic it will register a new ticker with `0 value for registeration date & expiryDate`.

satyamakgec (changes requested):

satyamakgec:
Also did a small commit to optimize more.

contracts/SecurityTokenRegistry.sol
adamdossa: Ah sorry - this was a typo from when I was tracking down a test case failure... Fixing now.

contracts/SecurityTokenRegistry.sol
adamdossa: Hey @satyamakgec,

For the first comment (always deleting the ticker owner before recreating it) - basically it is more gas expensive, but reduces the code size, so it was a bit of a trade-off. I guess we will use this function rarely (possibly never) to modify ticker details (mainly used to add new tickers, which isn't inefficient) so the trade-off seemed reasonable.

For the second part, my thinking was that if we were "artificially" creating a new ticker, it would be better to set the expiry / registration dates to 0 rather than pretending that it had been registered non-artificially and setting them to non-0 values - also cleaned up the code and made the size smaller ;-). If we wanted to we could always create the ticker first (with specified expiry / registration dates) then setup the security token.

contracts/SecurityTokenRegistry.sol
adamdossa: I reverted these changes.

contracts/SecurityTokenRegistry.sol
adamdossa: @satyamakgec 

I changed the logic so that if you're creating a ticker through a call to `modifySecurityToken` then the registration / expiry times are set as `now` and `now + expiryLimit`.

satyamakgec (approved):

PR #297, opened 2018-10-01
Fix namespace squatting in MR

adamdossa:
1. You need to be `owner` to register (and verify) modules until `customModulesAllowed` is switched on in the feature registry.

2. Cleaned getters for MR - we now have:
```
    function getTagsByTypeAndToken(uint8 _moduleType, address _securityToken) external view returns(bytes32[], address[]);
    function getTagsByType(uint8 _moduleType) external view returns(bytes32[], address[]);
    function getModulesByType(uint8 _moduleType) external view returns(address[]);
    function getModulesByTypeAndToken(uint8 _moduleType, address _securityToken) external view returns (address[]);

```
3. Removed - tags looked up directly in module factories:
`addTagsByModuleType`
`removeTagsByModuleType`


contracts/ModuleRegistry.sol
```diff
@@ -149,6 +147,9 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
      * @param _moduleFactory is the address of the module factory to be registered
      */
     function registerModule(address _moduleFactory) external whenNotPausedOrOwner {
+        if (IFeatureRegistry(getAddress(Encoder.getKey('featureRegistry'))).getFeatureStatus("customModulesAllowed")) {
+            require(msg.sender == getAddress(Encoder.getKey("owner")), "Only owner allowed to register modules");
```
pabloruiz55: @adamdossa this should also say `|| IOwnable(_moduleFactory).owner() == IOwnable(msg.sender).owner()`
so even when the "customModulesAllowed" is set, Polymath can still register someone else's module on their behalf, if for some reason we wanted so.

pabloruiz55: Actually, the require is wrong, it should be:

```
require(msg.sender == IOwnable(_moduleFactory).owner() || msg.sender == getAddress(Encoder.getKey('owner')),
        "msg.sender must be the Module Factory owner or registry curator");
```

pabloruiz55:
Looks good now, @satyamakgec please review and approve if everything looks ok.

satyamakgec (approved):

PR #299, opened 2018-10-02
Dividend name

### LINKED ISSUE #295: Add string identifiers to Ether & ERC20 dividends
We would need to have a `name` / `title` field on the dividends struct so they are better identifiable.

    struct Dividend {
      uint256 checkpointId;
      string name;
      ...
    }

When a dividend is created it should also require a name to be passed and event should emit that as well.

Relates to the `EtherDividendCheckpoint` and `ERC20DividendCheckpoint` modules.

maxsam4:

Enhancement

Dividends are unnamed

### What is the new behavior?
Dividend have a name

Need to pass Dividend name when creating a Dividend.

### Any Other information:
1) Apparently Sublime added whitespace changes to the whole contracts :/. Add ?w=1 to the URL to see the diff with whitespace ignored (https://github.com/PolymathNetwork/polymath-core/pull/299/commits/213b318984176aedecc31e624c88c13cad913f45?w=1).
2) I made _createDividendWithCheckpointAndExclusions an internal function to save gas when being called from public functions. public -> public function calls require all the parameters to be copied into memory again while public -> internal will just pass on the reference.
3) I am unable to run tests on my local machine for now so using travis for testing.

Fixes #295


adamdossa:
@maxsam4 Looks good to me - only comment is please could the function:  
`_CallERC20DividendDepositedEvent`
be renamed to:
`_emitERC20DividendDepositedEvent`

contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
```diff
@@ -1,157 +1,205 @@
...
+        address[] _excluded,
+        string _name
+    ) 
+        internal  
+    {
+        require(_excluded.length <= EXCLUDED_ADDRESS_LIMIT, "Too many addresses excluded");
+        require(_expiry > _maturity, "Expiry is before maturity");
+        require(_expiry > now, "Expiry is in the past");
+        require(_amount > 0, "No dividend sent");
+        require(_token != address(0), "0x not valid token");
```
satyamakgec, line 119: add the sanity check on the `_name`. `require(bytes(_name).length > 0);`

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -1,147 +1,182 @@
...
+        uint256 _checkpointId, 
+        address[] _excluded, 
+        string _name
+    ) 
+        internal
+    {
+        require(_excluded.length <= EXCLUDED_ADDRESS_LIMIT, "Too many addresses excluded");
+        require(_expiry > _maturity, "Expiry is before maturity");
+        require(_expiry > now, "Expiry is in the past");
+        require(msg.value > 0, "No dividend sent");
```
satyamakgec, line 102: same sanity check needed here

satyamakgec (changes requested):
I am not a big fan of `bytes32` but if we can replace the datatype to bytes32 for `name` then we can `indexed` it in the events that may useful for the dApp team.

contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
maxsam4: do we need to make _name a required field?

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
maxsam4: Added

contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
maxsam4: Changes done

pabloruiz55 (approved):

pabloruiz55:
looks good! Needs @satyamakgec 's approval

satyamakgec (approved):

PR #304, opened 2018-10-03
Clean up test cases

maxsam4:

1) Added helper for catching exceptions in test cases/
2) @satyamakgec Added poly initializer helper
3) Ran prettier to make code more readable.

NO

### Any Other information:
Work has just been started on it. In conjunction with @satyamakgec 

UPDATE
I think we should merge this branch immediately AFTER code freeze so that the new modules can make use of new helpers.
We'll continue working on adding test cases and improving coverage in a new PR.


contracts/modules/Checkpoint/DividendCheckpoint.sol
```diff
@@ -50,9 +50,9 @@ contract DividendCheckpoint is ICheckpoint, Module {
 
     modifier validDividendIndex(uint256 _dividendIndex) {
         require(_dividendIndex < dividends.length, "Incorrect dividend index");
+        require(!dividends[_dividendIndex].reclaimed, "Dividend has been reclaimed by issuer");
         require(now >= dividends[_dividendIndex].maturity, "Dividend maturity is in the future");
         require(now < dividends[_dividendIndex].expiry, "Dividend expiry is in the past");
-        require(!dividends[_dividendIndex].reclaimed, "Dividend has been reclaimed by issuer");
         _;
     }
```
maxsam4, line 58: This was done to cover the remaining branch as dividend can not be reclaimed before expiry is over. `require(!dividends[_dividendIndex].reclaimed)` is redundant but maybe required later if a module allows reclaiming of dividends before expiry.

maxsam4: The comment preview does not show it but I have actually moved the require statement above rather than removing.

test/helpers/exceptions.js
```diff
@@ -0,0 +1,44 @@
...
+    assert(
+    error.message.startsWith(PREFIX + message),
+    "Expected an error starting with '" +
+        PREFIX +
+        message +
+        "' but got '" +
+        error.message +
+        "' instead"
+    );
+    console.log(reason);
```
maxsam4: I added the reason string because we were printing reason for every fail tx. More I use it, more I think that it's redundant. the content inside it("") is the same as this reason. Should we remove this reason and use proper describes and its?

maxsam4:
Checkpoint modules now have 100% coverage except the 3 require(token.transfer) branches. We'll need to create a mock token that returns false on transfer to cover those.

contracts/modules/Checkpoint/DividendCheckpoint.sol
adamdossa: Makes sense. Since this is a base class let's leave the logic as you suggested even though in the current inheriting classes it is redundant (as you can't reclaim a dividend and still be before the expiry).

test/helpers/exceptions.js
adamdossa: @satyamakgec to comment here.

adamdossa (approved):

satyamakgec:
Not merge this PR. I am working on test cases Improvement.

test/helpers/exceptions.js
satyamakgec: Yes, @maxsam4 we can remove the reason string. It is same as `it` statement string.

maxsam4:
Do't worry @satyamakgec . I am not done either. Party has just started :)

maxsam4:
Created a script to replace all current try catch blocks with catchRevert(). A lot of stuff must have broke and will need fixing. Gonna continue work tomorrow.

I have also pushed the script "ex.js" if anyone wanna play around. (to be deleted before merging)

maxsam4:
Instrumenting timed out without deepSkip. Works fine on local system though.
We can do two things
1) Increase wait_time in travis
2) Enable deepSkip and not skip oracles and mocks in coverage. (Skipping them with deepSkip gives error)

I am gonna go with the second method for now.

PR #305, opened 2018-10-03
Support multiple module types

adamdossa:
Support modules having multiple types.

To Do: Add proof-of-concept module for testing multiple types

Includes lots of refactoring in SecurityToken to manage contract size limitations.

@VictorVicente I checked the CLI code and couldn't see any impact.


VictorVicente:
Agreed. It doesn't affect CLI.

contracts/ModuleRegistry.sol
```diff
@@ -155,8 +155,18 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
         }
         require(getUint(Encoder.getKey('registry', _moduleFactory)) == 0, "Module factory should not be pre-registered");
         IModuleFactory moduleFactory = IModuleFactory(_moduleFactory);
-        uint8 moduleType = moduleFactory.getType();
-        require(moduleType != 0, "Factory moduleType should not equal to 0");
+        //Enforce type uniqueness
+        uint256 i;
+        uint256 j;
+        uint8[] memory moduleTypes = moduleFactory.getTypes();
+        for (i = 0; i < moduleTypes.length; i++) {
```
maxsam4: We can start the loop from `i = 1`

contracts/libraries/TokenLib.sol
```diff
@@ -0,0 +1,152 @@
...
+            _checkpoints.push(
+                TokenLib.Checkpoint({
+                    checkpointId: _currentCheckpointId,
+                    value: _newValue
+                })
+            );
+            return;
+        }
+        //No new checkpoints since last update
+        if (_checkpoints[_checkpoints.length - 1].checkpointId == _currentCheckpointId) {
```
maxsam4: `if (_checkpoints[_checkpoints.length - 1].checkpointId == _currentCheckpointId)` 
can be converted to 
`if (_checkpoints.length > 0 && _checkpoints[_checkpoints.length - 1].checkpointId == _currentCheckpointId)`

Then we can completely remove the `if (_checkpoints.length == 0)` case. 
It should bring down the size by tiny a bit.

contracts/interfaces/ISecurityToken.sol
```diff
@@ -33,14 +33,14 @@ interface ISecurityToken {
      * @notice Burn function used to burn the securityToken
      * @param _value No. of tokens that get burned
      */
-    function burn(uint256 _value) external returns (bool success);
+    function burn(uint256 _value) external;
```
maxsam4, line 36: I am not sure if we should remove return from _standard_ functions.

contracts/modules/Checkpoint/DividendCheckpoint.sol
```diff
@@ -131,9 +131,10 @@ contract DividendCheckpoint is ICheckpoint, Module {
      */
     function pushDividendPayment(uint256 _dividendIndex, uint256 _start, uint256 _iterations) public withPerm(DISTRIBUTE) validDividendIndex(_dividendIndex) {
         Dividend storage dividend = dividends[_dividendIndex];
-        uint256 numberInvestors = ISecurityToken(securityToken).getInvestorsLength();
+        address[] memory investors = ISecurityToken(securityToken).getInvestors();
+        uint256 numberInvestors = investors.length;
         for (uint256 i = _start; i < Math.min256(numberInvestors, _start.add(_iterations)); i++) {
```
maxsam4: we can do 
`uint256 numberInvestors = Math.min256(investors.length, _start.add(_iterations))`
and then in the loop, just compare `i` with `numberInvestors`
It will skip the need of executing `min256` and `add` function on every iteration.

contracts/tokens/SecurityToken.sol
```diff
@@ -217,102 +210,106 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-        modules[moduleType].push(module);
+        uint256[] memory moduleIndexes = new uint256[](moduleTypes.length);
+        uint256 i;
+        for (i = 0; i < moduleTypes.length; i++) {
+            moduleIndexes[i] = modules[moduleTypes[i]].length;
+        }
+        modulesToData[module] = TokenLib.ModuleData(moduleName, module, _moduleFactory, false, moduleTypes, moduleIndexes, names[moduleName].length);
+        for (i = 0; i < moduleTypes.length; i++) {
+            /* modulesToData[module].moduleType[moduleTypes[i]] = true; */
+            modules[moduleTypes[i]].push(module);
```
maxsam4, line 241: I think `modules[moduleTypes[i]].push(module);` can be moved to the loop above and we can remove this extra loop.

contracts/tokens/SecurityToken.sol
```diff
@@ -217,102 +210,106 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
     function unarchiveModule(address _module) external onlyOwner {
-        require(modulesToData[_module].isArchived, "Module already unarchived");
-        emit ModuleUnarchived(modulesToData[_module].moduleType, _module, now);
-        modulesToData[_module].isArchived = false;
+        TokenLib.unarchiveModule(modulesToData[_module], _module);
+    }
+
+    function _removeModuleWithIndex(uint8 _type, uint256 _index) internal {
+        uint256 length = modules[_type].length;
+        modules[_type][_index] = modules[_type][length - 1];
```
maxsam4: `modules[_type][_index] = modules[_type][length - 1];` can be moved to the if block below and length-- can be done after the if statement. Very minor change that can save some gas when `length - 1 == index`.

contracts/tokens/SecurityToken.sol
```diff
@@ -811,16 +715,19 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
         emit ForceTransfer(msg.sender, _from, _to, _value, verified, _data);
         emit Transfer(_from, _to, _value);
-        return true;
+    }
+
+    function _transfer(address _from, address _to, uint256 _value) internal returns(bool) {
+        require(_value <= balances[_from]);
+        bool verified = _updateTransfer(_from, _to, _value);
+        balances[_from] = balances[_from].sub(_value);
+        return verified;
```
maxsam4: maybe add `_value` to `balances[_to]`  here and use `_burn()` in `forceBurn()` to remove ambiguity.

maxsam4 (commented):
The logic seems sound to me but I don't really know the exact requirements. I have listed some minor possible optimizations below. @satyamakgec and @pabloruiz55 can take over from here.

contracts/interfaces/ISecurityToken.sol
satyamakgec: agree with @maxsam4

contracts/libraries/TokenLib.sol
```diff
@@ -0,0 +1,152 @@
...
+    // Struct for module data
+    struct ModuleData {
+        bytes32 name;
+        address module;
+        address moduleFactory;
+        bool isArchived;
+        uint8[] moduleTypes;
+        uint256[] moduleIndexes;
+        uint256 nameIndex;
+        mapping (uint8 => uint256) moduleIndex;
```
satyamakgec: We are not using this mapping anywhere. Can we remove it?

satyamakgec:
Logic is good @adamdossa but we can also move the investor related functions into the library. It does make sense to move them as not many other functions depend on it. It will save up to `.30 KB` and also doesn't introduce too much complexity in the code.

I did some changes regarding that check here https://github.com/PolymathNetwork/polymath-core/blob/some-changes-adampr/contracts/libraries/TokenLib.sol

If you think that these changes are worth to add then we can merge these changes into yours.

maxsam4:
Just some handy stats.
Gas used for deploying ST (this branch):  7411740
Gas used for deploying ST (satyam's branch):  7378646

contracts/tokens/SecurityToken.sol
adamdossa: I refactored this slightly differently, but still used _burn under the covers. Please check @maxsam4

maxsam4: Looks good to me @adamdossa :+1:

contracts/interfaces/ISecurityToken.sol
adamdossa: Since burn (and burnFrom) can only be used from a module, I am going to leave without the return statement. IMO the best semantics for these types of operations should be that they revert on failure. transfer / transferFrom are exceptions in order to meet ERC20 spec.

satyamakgec (approved):

PR #314, opened 2018-10-04
Migrate the investors logic from ST to library

satyamakgec:


maxsam4:
Maybe we should remove `function investorCount()` from ISecurityToken interface.

contracts/interfaces/ISecurityToken.sol
```diff
@@ -245,4 +245,9 @@ interface ISecurityToken {
      * @notice Use to get the version of the securityToken
      */
      function getVersion() external view returns(uint8[]);
+
+     /**
+     * @notice gets the investor count
+     */
+     function getInvestorCount() external view returns(uint256)
```
maxsam4: We should probably remove `investorCount()` in favor of `getInvestorCount()`

maxsam4 (approved):

PR #315, opened 2018-10-04
Add additional transfer / mint / burn functions with _data parameter

adamdossa:
This PR adds transfer, mint & burn functions with an additional `bytes _data` parameter which is passed onto TransferManagers.

This would allow us to do off-chain whitelisting - for example if an exchange is generating a dynamic deposit address for a customer they could also provide a signed piece of data which authorises this address (assuming the issuer had added a TM that accepted this signed data).

More or less inspired from ERC 1400 although more ERC20 like in naming.

For the burn functions, I didn't include the non-data versions as these aren't part of the ERC20 specification in any case (probably the same argument could be made for mint, but that is already used in lots of places so I left the old version of the function).

### DIFF
--- a/contracts/interfaces/ISecurityToken.sol
+++ b/contracts/interfaces/ISecurityToken.sol
@@ -30,17 +30,28 @@ interface ISecurityToken {
     function mint(address _investor, uint256 _value) external returns (bool success);
 
     /**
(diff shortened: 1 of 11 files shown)


satyamakgec (approved):
Looks good to me @pabloruiz55 please have a second look

PR #319, opened 2018-10-05
Migrate fix

maxsam4:
1) `npm i fs`
2) save private key in ./privKey
3) `truffle migrate --network kovan`


maxsam4:
Nevermind
`Transaction gas price is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.`

satyamakgec:
Why would we need that change? @maxsam4

maxsam4:
@satyamakgec Because migration to Kovan are failing. Please refer to the slack channel for more info.

maxsam4:
Build is passing, docs are failing. 
Coverage is failing though on all pr since morning.

pabloruiz55:
Great job @maxsam4 ! I was able to test this and I can confirm I was able to deploy without any issues.

Issue #267, opened 2018-09-19
Module Development Task: Volume Restriction (Maximum Volume Transfer)


CPSTL:
# Module Development Task
# Volume Restriction  (Maximum Volume Transfer)

Module Type | Bounty | Completion deadline
-- | -- | --
Transfer Manager | Approx. 10, 000 POLY | 21 days to complete once task is confirmed by external developer(s).

# Short Summary: 

For this module, ‘Maximum Volume Transfer’ we want to implement it in a way that you can only transfer a limited number of the total number of tokens within a certain time period.

Any  token-holder can only transfer up to a maximum of 500,000 tokens in a 90 day period, regardless of the other restrictions (0.5% of the 100,000,000 tokens issued by the company). This applies to all token-holders and not just affiliated parties.

**Note:** This can be implemented as TM module with current architecture.

# Bounty Requirements
1. Module specs are fulfilled
2. Module is implemented using appropriate module interface
3. Module is tested with >95% branch coverage
4. Module is delivered before deadline
5. Module does not present any bugs.
6. 21 days to complete module from developers initial start date.

**IMPORTANT:** Please work from the development-1.5.0 branch instead of master.
Implementation Suggestion

# Module Details Spec (Wireframe) 
**Wireframe Link:** https://www.lucidchart.com/invitations/accept/75bdef55-cd39-48d9-bc1c-ac1d2fc76a39

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 10000.0 POLY (1499.0 USD @ $0.15/POLY)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/PolymathNetwork/polymath-core/267/1256).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/PolymathNetwork/polymath-core/267/1256).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $26,152.48 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. Started 3. Submitted 4. **Done** 

<hr>

__Work has been started__.


[...]

PR #320, opened 2018-10-05
Maximum Volume Restrictions Manager

vyomshm:

feature - ref #267 

#267 
No way to add global token volume transfer restrictions

### What is the new behavior?
Feature added - A new Transfer Manager module
Issuers can now create(as well as remove and modify) multiple maximum volume transfer restrictions which will be applied globally to every transfer

No


satyamakgec:
Hey, @vyomshm let me know when should I start reviewing your code?

vyomshm:
Hey @satyamakgec Please review my PR

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        RollingIntervals rollingPeriodInterval;
+    }
+
+    MaximumVolumeRestriction[] public maximumVolumeRestrictions;
+
+    // mapping from address to RollingIntervals to periodIds to tokens transfered 
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) volumeTally;
+
+    // Emit whenever a new maximum volume restriction gets added
+    event MaximumVolumeRestrictionAdded(
```
satyamakgec: rename it to `AddNewMaximumVolumeRestriction`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        address indexed from,
+        uint256 timestamp,
+        uint256 _maximumVolume, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
+    );
+
+    // Emit whenever a new maximum volume restriction gets modified
+    event MaximumVolumeRestrictionModified(
```
satyamakgec: rename it to `ModifyMaximumVolumeRestriction`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        uint256 timestamp,
+        uint256 idx, 
+        uint256 _maximumVolume, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
+    );
+
+    // Emit whenever a new maximum volume restriction gets removed
+    event MaximumVolumeRestrictionRemoved(
```
satyamakgec: rename it to `RemoveMaximumVolumeRestricion`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * @notice This function returns the signature of configure function
+     */
+    function getInitFunction() public pure returns (bytes4) {
+        return bytes4(0);
+    }
+
+    /// @notice Used to verify the transfer transaction according to the rule implemented in the trnasfer managers
```
satyamakgec: add a descriptive natspec comments

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    /**
+    * @notice allows admin to modify a maximum volume transfer restriction
+    * @param idx representing the array index of the maximum volume transfer restriction
+    * @param _maximumVolume maximum token volume that can be transfered
+    * @param _startTime timestamp representing when this maximum volume restriction starts
+    * @param _endTime timestamp representing when this maximum volume restriction ends
+    * @param _rollingPeriodInterval interval period representing one of five RollingIntervals
+    */
+    function modifyMaximumTransferRestriction(
+        uint256 idx,
```
satyamakgec: rename it to `_index`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        if(_rollingPeriodInterval == RollingIntervals.Daily){
+            period = 24 hours;
+        }else if(_rollingPeriodInterval == RollingIntervals.Weekly){
+            period = 168 hours;
+        }else if(_rollingPeriodInterval == RollingIntervals.Quarterly){
+            period = 2920 hours;
+        }else if(_rollingPeriodInterval == RollingIntervals.Biannually){
+            period = 4380 hours;
+        }else{
+            period = 8760 hours;
```
satyamakgec: change it to  365 * 24 hours that makes easy to read.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    function verifyTransfer(address _from, address /*_to*/, uint256 _amount, bytes /* _data */, bool _isTransfer) public returns(Result) {
+        // function must only be called by the associated security token if _isTransfer == true
+        require(_isTransfer == false || msg.sender == securityToken, "Sender is not owner");
+
+        if (!paused) {
+            for(uint256 i = 0; i < maximumVolumeRestrictions.length; i++){
+                MaximumVolumeRestriction memory restriction = maximumVolumeRestrictions[i];
+                if(now >= restriction.startTime && now <= restriction.endTime){
+                    uint256 period = _getPeriod(restriction.rollingPeriodInterval);
+                    uint256 periodId = now.div(period);
```
satyamakgec: periodId should be `(now - startTime).div(period)`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    MaximumVolumeRestriction[] public maximumVolumeRestrictions;
+
+    // mapping from address to RollingIntervals to periodIds to tokens transfered 
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) volumeTally;
+
+    // Emit whenever a new maximum volume restriction gets added
+    event MaximumVolumeRestrictionAdded(
+        address indexed from,
```
satyamakgec: no need of `from` address. it always be owner or a delegate

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        uint256 timestamp,
+        uint256 _maximumVolume, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
+    );
+
+    // Emit whenever a new maximum volume restriction gets modified
+    event MaximumVolumeRestrictionModified(
+        address indexed from,
```
satyamakgec: no need of `from` address. it always be owner or a delegate

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+        uint256 timestamp,
+        uint256 idx,
```
satyamakgec: change name `idx` to `index`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+    * @notice allows admin to add multiple new maximum volume transfer restrictions
+    * @param _maximumVolumes maximum token volumes that can be transfered
+    * @param _startTimes timestamps representing when these maximum volume restrictions start
+    * @param _endTimes timestamps representing when these maximum volume restrictions end
+    * @param _rollingPeriodIntervals interval periods representing one of five RollingIntervals
+    */
+    function addMultipleMaximumVolumeRestrictions(
```
satyamakgec: rename it to `addMaxVolumeRestricionsMulti()`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+    * @notice allows admin to add new maximum volume transfer restrictions
+    * @param _maximumVolume maximum token volume that can be transfered
+    * @param _startTime timestamp representing when this maximum volume restriction starts
+    * @param _endTime timestamp representing when this maximum volume restriction ends
+    * @param _rollingPeriodInterval interval period representing one of five RollingIntervals
+    */
+    function addMaximumVolumeRestriction(
```
satyamakgec: rename it to `addMaxVolumeRestricion`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) volumeTally;
+
+    // Emit whenever a new maximum volume restriction gets added
+    event MaximumVolumeRestrictionAdded(
+        address indexed from,
+        uint256 timestamp,
+        uint256 _maximumVolume, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
```
satyamakgec, line 35: emit the `index` of the new volume restriction

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,290 @@
...
+            for(uint256 i = 0; i < maximumVolumeRestrictions.length; i++){
+                MaximumVolumeRestriction memory restriction = maximumVolumeRestrictions[i];
+                if(now >= restriction.startTime && now <= restriction.endTime){
+                    uint256 period = _getPeriod(restriction.rollingPeriodInterval);
+                    uint256 periodId = now.div(period);
+                    if(_amount.add(volumeTally[_from][uint256(restriction.rollingPeriodInterval)][periodId]) > restriction.maximumVolumeAllowed){
+                        return Result.INVALID;
+                    }else{
+                        volumeTally[_from][uint256(restriction.rollingPeriodInterval)][periodId] = _amount.add(volumeTally[_from][uint256(restriction.rollingPeriodInterval)][periodId]);
+                    }
```
satyamakgec, line 95: Add some description on your algorithm to understand the code steps

satyamakgec (changes requested):

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
F-OBrien: Also is the number of hours for the RollingIntervals.Quarterly correct? 365*24/4 = 2190 hours not 2920

satyamakgec: @FOBCrypto Thanks for pointing out. Yes, it is miscalculated but this calculation get resolved if he put the hours in factors

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: done!

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: done. adding an `_index` param to be the first

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: Fixed it! @FOBCrypto thanks!

vyomshm:
@satyamakgec @FOBCrypto I've made the suggested changes, please review again.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+
+import "./ITransferManager.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+/**
+ * @title Transfer Manager module for capping the amount of token transfers within a period
+ */
+contract MaximumVolumeTransferManager is ITransferManager {
+	
+	using SafeMath for uint256;
```
maxsam4: Linting.
```suggestion
    using SafeMath for uint256;
```

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
+    ) internal view {
+        require(
+            uint256(RollingIntervals.Annually) >= _rollingPeriodInterval,
+            "invalid rolling period interval - should be < 5"
+        );
+        require(_startTime == 0 || _startTime >= now, 'invalid startTime');
+        require(_endTime > 0 && _endTime > _startTime, 'invalid endTime');
```
maxsam4: `_endTime > 0` is not required as it will always be true when `_endTime > _startTime`.
```suggestion
        require(_endTime > _startTime, 'invalid endTime');
```

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+    function _checkRestrictionParams(
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval
+    ) internal view {
+        require(
+            uint256(RollingIntervals.Annually) >= _rollingPeriodInterval,
+            "invalid rolling period interval - should be < 5"
+        );
+        require(_startTime == 0 || _startTime >= now, 'invalid startTime');
```
maxsam4: StartTime 0 should not be allowed. `_startTime < now` should be allowed though for editing/adding restriction that have already started.
```suggestion
        require(_startTime != 0, 'invalid startTime');
```

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint256 _maximumVolume, 
+        uint256 _startTime, 
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval) 
+    public withPerm(ADMIN) {
+        _checkRestrictionParams(_startTime, _endTime, _rollingPeriodInterval);
+
+        if(_startTime == 0) {
+            _startTime = now;
+            require(_endTime > _startTime, "end time must be greater than start time");
```
maxsam4: No need for this require statement.
Just move the if block above `_checkRestrictionParams()`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        require(_maximumVolumes.length >= 1, "No restrictions provided");
+        require(
+            _maximumVolumes.length == _startTimes.length &&
+            _maximumVolumes.length == _endTimes.length &&
+            _maximumVolumes.length == _rollingPeriodIntervals.length,
+            "Input array length mismatch"
+        );
+
+        for(uint256 i = 0; i < _maximumVolumes.length; i++){
+            addMaxVolumeRestricion(
```
maxsam4: Create an internal function and use it here as well as in `addMaxVolumeRestricion()`. Doing so will save gas as parameters are not copied into memory again for internal functions.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint256 _endTime, 
+        uint256 _rollingPeriodInterval) 
+    public withPerm(ADMIN) {
+        require(_index < maximumVolumeRestrictions.length, "invalid index");
+        require(now < maximumVolumeRestrictions[_index].endTime, "restriction is no longer active");
+        _checkRestrictionParams(_startTime, _endTime, _rollingPeriodInterval);
+
+        if(_startTime == 0) {
+            _startTime = now;
+            require(_endTime > _startTime, "end time must be greater than start time");
```
maxsam4: Remove the require statement, move the if block above `_checkRestrictionParams()`

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+            _rollingPeriodInterval
+        );
+    }
+
+    /**
+    * @notice allows admin to delete a maximum volume transfer restriction
+    * @param _index representing the array index of the maximum volume transfer restriction
+    */
+    function removeMaximumTransferRestriction(uint256 _index) public withPerm(ADMIN){
+        require (maximumVolumeRestrictions.length != 0, "no existing restrictions");
```
maxsam4: Not required. The second statement covers this as well.
```suggestion
```

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        require (maximumVolumeRestrictions.length > _index, "invalid index");
+
+        if(_index < (maximumVolumeRestrictions.length - 1)){
+            MaximumVolumeRestriction memory lastRestriction = maximumVolumeRestrictions[maximumVolumeRestrictions.length - 1];
+            maximumVolumeRestrictions[_index] = lastRestriction;
```
maxsam4: No need to copy it in memory first. Just copy directly from storage.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        maximumVolumeRestrictions.length = (maximumVolumeRestrictions.length).sub(1);
```
maxsam4: SafeMath is not required here as it will revert before reaching this statement if `maximumVolumeRestrictions.length == 0`
```suggestion
        maximumVolumeRestrictions.length = maximumVolumeRestrictions.length - 1;
```

scripts/test.sh
```diff
@@ -63,7 +63,7 @@ start_testrpc() {
   if ! [ -z "${TRAVIS_PULL_REQUEST+x}" ] && [ "$TRAVIS_PULL_REQUEST" != false ]; then
     node_modules/.bin/testrpc-sc --gasLimit 0xfffffffffff --port "$testrpc_port" "${accounts[@]}" > /dev/null &
   else
-    node_modules/.bin/ganache-cli --gasLimit 8000000 "${accounts[@]}" > /dev/null &
+    node_modules/.bin/ganache-cli --gasLimit 0xfffffffffff "${accounts[@]}" > /dev/null &
```
maxsam4: We need to keep our tests within the limit of 8m gas (maximum allowed on main net)
```suggestion
    node_modules/.bin/ganache-cli --gasLimit 8000000 "${accounts[@]}" > /dev/null &
```

maxsam4 (changes requested):
Other than the suggested changes, Can you please also increase the coverage of your code?
Branch coverage of `MaximumVolumeTransferManager.sol` is at 80% currently. We would ideally like to see 100% coverage but a minimum of 95% is required.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: okay but accordingly I've added `_endTime > now` to prevent anyone from modifying expired restrictions or adding invalid ones.
```suggestion
        require(_startTime == 0 || _startTime >= now, 'invalid startTime');
        require(_endTime > _startTime && _endTime > now, 'invalid endTime');
```

maxsam4:
@vyomshm Please push the changes before marking conversations as resolved.

vyomshm:
@satyamakgec @maxsam4 @FOBCrypto I've increased my test coverage + implemented changes as discussed.  Ready for another review👍

maxsam4:
Nice work @vyomshm !

The branch coverage is still at 92.5% with 3 uncovered branches. I would love for them to be covered as well (all of them can be covered). You can see the uncovered branches at https://coveralls.io/builds/19635412/source?filename=contracts%2Fmodules%2FTransferManager%2FMaximumVolumeTransferManager.sol

vyomshm:
@maxsam4 @satyamakgec @FOBCrypto Coverage is now at 100%

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -134,7 +134,6 @@ contract MaximumVolumeTransferManager is ITransferManager {
         uint256[] _endTimes, 
         uint256[] _rollingPeriodIntervals) 
     public withPerm(ADMIN) {
-        require(_maximumVolumes.length >= 1, "No restrictions provided");
```
maxsam4: Put the sanity check back.
Add a test case with empty arrays to cover the branch.

maxsam4 (changes requested):

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: done!

vyomshm:
@satyamakgec @FOBCrypto @maxsam4 made the changes. please have a look

vyomshm:
@maxsam4 have you had a chance to look through this?  Would love to see this get resolved before you guys get too busy with pre-devcon4 events  :)

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,307 @@
...
+            _endTime,
+            _rollingPeriodInterval
+        );
+    }
+
+    /**
+    * @notice allows admin to delete a maximum volume transfer restriction
+    * @param _index representing the array index of the maximum volume transfer restriction
+    */
+    function removeMaximumTransferRestriction(uint256 _index) public withPerm(ADMIN){
```
maxsam4, line 201: Can you please add a batch/multi versions of this.

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+    * @notice allows admin to modify a maximum volume transfer restriction
+    * @param _index representing the array index of the maximum volume transfer restriction
+    * @param _maximumVolume maximum token volume that can be transfered
+    * @param _startTime timestamp representing when this maximum volume restriction starts
+    * @param _endTime timestamp representing when this maximum volume restriction ends
+    * @param _rollingPeriodInterval interval period representing one of five RollingIntervals
+    */
+    function modifyMaximumTransferRestriction(
```
maxsam4, line 163: Can you please add a batch/multi versions of this.

maxsam4 (changes requested):

contracts/modules/TransferManager/MaximumVolumeTransferManager.sol
vyomshm: @satyamakgec  @maxsam4 Currently, removing a single array element will almost always change the order of restrictions in the array( except when the restriction to be deleted in the last one). This could render some of the indices provided unusable and therefore cause the function to revert.

I could write an implementation which would preserve the order of restrictions by shifting the elements instead of swapping them(as implemented now) but doing so would make the function extremely costly in terms of gas. 
```
string element = myArray[index];
for (uint i = index; i < myArray.length - 1; i++) {
  myArray[index] = myArray[index + 1];
}
myArray.length--;
```

Please advise on how to proceed with this.

maxsam4: @vyomshm Can you update the indices as well when deleting an array element?

maxsam4: I think one way around the problem can be by adding a unique id to `MaximumVolumeRestriction` struct and maintaining an array of that id for parsing. `maximumVolumeRestrictions` can be made a mapping rather than an array then.

maxsam4 (changes requested):
I was going through the logic once again and it seems flawed. There can be multiple restrictions of overlapping periods which can make the logic fail. For example, consider a case when there is a monthly restriction started on 1st Jan and another monthly restriction started on 15th Jan. If anyone tries to transfer tokens on 16th of Jan, the below line will be executed twice
`volumeTally[_from][monthly][0] = _amount.add(volumeTally[_from][monthly][0]);`
This should have been executed only once.

There are a lot more problematic scenarios due to this. Another example with the same restrictions will be when a transfer is done on 1st Feb.  `volumeTally[_from][monthly][0]` will now contain info about the volume of trade that happened on 1st of Jan but the second restriction will not be expecting that.

Basically, the logic will fail when there are overlapping trade restriction periods. Please, rethink about it and try to fix it.

maxsam4:
I was just going through the requirements and specs of this module and seems like your implementation and our requirements don't completely match. Please have a look at https://www.lucidchart.com/invitations/accept/75bdef55-cd39-48d9-bc1c-ac1d2fc76a39 again.

We want different people to be linked with different volume restrictions. Not all restrictions need to be imposed on every user.

vyomshm:
@maxsam4 With regards to your first comment, `periodId` in both cases would not be the same as it is dependent on `startTime` of a restriction  and defined as 
`uint256 periodId = (now.sub(restriction.startTime)).div(period);` 
therefore, the line 
`volumeTally[_from][uint256(restriction.rollingPeriodInterval)][periodId] = _amount.add(volumeTally[_from][uint256(restriction.rollingPeriodInterval)][periodId]);`

would be executed only when relevant.

vyomshm:
`This applies to all token-holders and not just affiliated parties.`
@maxsam4 I'll admit that from this line in "Short summary" section of #267  as posted by @CPSTL  as well as the absence of any `address` field in the spec diagram, I **inferred that all max volume transfer restrictions were supposed to be globally imposed on all token holders and not just specific addresses**
screenshot 2018-10-27 at 12 34 39 am

Please clarify if my interpretation of the module is wrong, so I can rectify my codebase.

maxsam4:
> @maxsam4 With regards to your first comment, `periodId` in both cases would not be the same as it is dependent on `startTime` of a restriction and defined as
> `uint256 periodId = (now.sub(restriction.startTime)).div(period);`
> would be executed only when relevant.

@vyomshm Solidity does not support decimals which means after `div(period)` we will have the same `periodId` for a lot of restrictions. All restrictions will start with periodId 0 and keep incrementing as time passes. Please let me know if I am understanding it wrong.



> Please clarify if my interpretation of the module is wrong, so I can rectify my codebase.

I agree that we could have been clearer about the specs but as the second assumption in lucidchart states, the token owner should be able to impose different restrictions on different sets of token holders. Please rework your approach accordingly. 
Thank you for cooperating.

vyomshm:
@CPSTL  @satyamakgec @maxsam4 Just to be clear, we want maxVolume transfer restrictions to exist for specific addresses only and not as Global impositions ?

This would require changing `addMaxVolumeRestricion` function signature to also include an `address` field. 
Also I could modify the `maximumVolumeRestrictions` to be a mapping instead of a global array like -
`mapping(address => MaximumVolumeRestriction) public maximumVolumeRestrictions`. This would also resolve the problem you described.

Can you confirm this is the approach  I should proceed with? 
I believe this would lead to an implementation that is quite similar as being taken for this issue -> https://github.com/PolymathNetwork/polymath-core/issues/268 .

maxsam4:
@vyomshm We want both the abilities. Global restrictions + restrictions on only a subset of address.
This module should cater to global restrictions and the other volume restriction module should cater to individual limits. 
The limits on a subset of investors can be catered to by both or any one of these modules. I'll have a chat with the team and get back to you regarding this.

vyomshm:
@maxsam4 okay, I'll wait.

Just so you know, I've figured out the easiest fix for the same periodId issue. As in my original implementation we should define periodId as follows-
`uint256 periodId = now.div(period);`

Also I believe we can implement the multi version of `removeMaximumTransferRestriction` by dynamically modifying the user provided indices according to elements already removed.

maxsam4:
> @maxsam4 okay, I'll wait.
> 
> Just so you know, I've figured out the easiest fix for the same periodId issue. As in my original implementation we should define periodId as follows-
> `uint256 periodId = now.div(period);`

That won't work for the same reasons I said when you first implemented it. Also, Even in this case, there will be collisions because solidity does not support floating points. It will round them to the nearest integer.

vyomshm:
@maxsam4 I understand what you're saying. As I see it the best way to avoid any sort of collisions will be to maintain a separate `volumeTally` for every restriction.

One way to do this would be to add to modify the `MaximumVolumeRestriction` struct to also include an `id`  and extending the multi-level `volumeTally` mapping another level.
```
uint256 internal _restrictionIdCount = 0;

struct MaximumVolumeRestriction{
        uint256 restrictionId;
        uint256 maximumVolumeAllowed;
        uint256 startTime;
        uint256 endTime;
        RollingIntervals rollingPeriodInterval;
}

// mapping from address to RollingIntervals to restrictionIds to periodIds to tokens transfered 
mapping(address => mapping(uint256 => mapping( uint256 => mapping(uint256 => uint256))) volumeTally;
```
`restrictionId` var could take its value from a global `restrictionIdCount` which would in turn be incremented by 1 every time a maximum volume restriction is created/modified. This would ensure that all `maxVolumeRestrictions` have a unique `restrictionId`

vyomshm:
@maxsam4 have you had a chance to talk with the team yet?

maxsam4:
@vyomshm Sorry, we are busy with devcon this week. I'll get back to you next week. Thanks for understanding.

vyomshm:
@maxsam4 Could you update me on how to proceed with this?

maxsam4:
Bounty paid. Thanks a lot for your contribution @vyomshm .

PR #321, opened 2018-10-05
Travis fix

maxsam4:
Debugging Travis PR coverage failures.


maxsam4:
Just an update:
Last commit narrows down the the problem to the new Single Trade Volume Restriction Manager Module. 
We know the problem is system dependent as it is working on my local machine but not on travis and other users' systems. 
We also know that the problem is with instrumenting in solidity coverage as not even my system is able to compile the instrumented contracts generated by @adamdossa 's system.

It might be that solc can not compile contracts of over a certain size. If that's the case, we should figure out why/how solidity coverage is able to generate smaller instrumented contracts on my system.

maxsam4:
Now this is interesting. No compilation error popped up but coverage still failed.

maxsam4:
Just an update:
commit 07d2444 shows that the error is in SingleTradeVolumeRestrictionManager.sol

.solcover.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
     norpc: true,
     port: 8545,
     copyPackages: ['openzeppelin-solidity'],
-    testCommand: 'node ../node_modules/.bin/truffle test `find test/*.js ! -name a_poly_oracle.js -and ! -name s_v130_to_v140_upgrade.js` --network coverage',
```
nervehammer: ` testCommand: 'node --max-old-space-size=4096 ../node_modules/.bin/truffle test `find test/*.js ! -name a_poly_oracle.js -and ! -name s_v130_to_v140_upgrade.js` --network coverage',`
`compileCommand: 'node --max-old-space-size=4096 ../node_modules/.bin/truffle compile --network coverage',`

nervehammer: @maxsam4 maybe this will help, i am not 100% sure but it worth a try :)

maxsam4: @nervehammer We need `find test/*.js ! -name a_poly_oracle.js -and ! -name s_v130_to_v140_upgrade.js`

`--max-old-space-size=4096` This however, might do the trick. Unlikely as no out of memory error is being thrown but worth a try at the very least. We are redirecting a lot of output to dev/null

maxsam4: Didn't work :(

maxsam4:
Two basic changes
1) Remove STVRM from coverage
2) Disable docs generation 

Commits can be squashed and merged to get back the green tick :)

pabloruiz55 (approved):

PR #327, opened 2018-10-09
Add a ScheduledCheckpoint PoC Module

adamdossa:
Proof of concept of scheduling future actions based on state updates.

ScheduledCheckpoint module is a Transfer & Checkpoint module. Every time there is a transfer it checks whether a new checkpoint needs to be created based on repeated schedules.


contracts/modules/Checkpoint/ScheduledCheckpoint.sol
```diff
@@ -0,0 +1,150 @@
...
+    function addSchedule(bytes32 _name, uint256 _startTime, uint256 _interval) external onlyOwner {
+        require(_startTime > now, "Start time must be in the future");
+        require(schedules[_name].name == bytes32(0), "Name already in use");
+        schedules[_name].name = _name;
+        schedules[_name].startTime = _startTime;
+        schedules[_name].nextTime = _startTime;
+        schedules[_name].interval = _interval;
+        schedules[_name].index = names.length;
+        names.push(_name);
+    }
```
satyamakgec: I think it is useful to add an event for the dApp. `emit addNewSchedule(..)`

contracts/modules/Checkpoint/ScheduledCheckpoint.sol
```diff
@@ -0,0 +1,150 @@
...
+    function removeSchedule(bytes32 _name) external onlyOwner {
+        require(schedules[_name].name == _name, "Name does not exist");
+        uint256 index = schedules[_name].index;
+        names[index] = names[names.length - 1];
+        names.length--;
+        if (index != names.length) {
+            schedules[names[index]].index = index;
+        }
+        delete schedules[_name];
+    }
```
satyamakgec: I think it is not worth to delete the data from the mapping because there is no another way to access the data once it is deleted from the contract (besides accessing the previous blocks). we can add a boolean flag to make it inactive.
I think dApp may need to show the old data as well if that schedule is not active.

satyamakgec (commented):
@adamdossa PR looks good to me. left two comments (concerns only). Though I don't play with library yet It takes more time in comparison to other contracts.

contracts/modules/Checkpoint/ScheduledCheckpoint.sol
```diff
@@ -0,0 +1,150 @@
...
+        bytes32 name;
+        uint256 startTime;
+        uint256 nextTime;
+        uint256 interval;
+        uint256 index;
+        uint256[] checkpointIds;
+        uint256[] timestamps;
+        uint256[] periods;
+    }
+
```
satyamakgec: one more thing we can do is compact binding in the struct if want to save some gas for the transaction.

contracts/modules/Checkpoint/ScheduledCheckpoint.sol
adamdossa: I think I will leave this as it is. This module is more designed as a proof of concept module rather than something that would be directly used. For a real module for this type of thing we would have specific product driven specs. which as you say may mean that we want to keep around previous "deleted" schedules.

contracts/modules/Checkpoint/ScheduledCheckpoint.sol
adamdossa: Yep - will leave this alone for now - a "real" module can make gas optimisations as needed. Just want this to be a kind of "how to use TMs to schedule future events" proof of concept ;-).

contracts/modules/Checkpoint/ScheduledCheckpoint.sol
satyamakgec: Make sense.

pabloruiz55 (approved):

PR #329, opened 2018-10-10
Voting module

comeonbuddy:
- [x ] The commit message follows our Submission guidelines
- [x ] Tests for the changes have been added (for bug fixes / features)
- [? ] Docs have been added / updated (for bug fixes / features)

upgrade the Weighted Voting module to 1.5.0
fixed all the broken functions. 
fixed necessary changes on test file.
Added new function to allow deactive ballot.

### What is the new behavior?

No


contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+        return true;
+    }
+
+    /**
+     * @notice Allows the token issuer to create a ballot
+     * @param _duration The duration of the voting period in seconds
+     * @return bool success
+     */
+    function createBallot(uint256 _duration) public onlyOwner returns (bool) {
+        uint256 ballotId = ballots.length;
```
adamdossa: Should check that _duration > 0 I think.

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @notice Allows the token issuer to create a ballot with custom settings
+     * @param _startTime Start time of the voting period in Unix Epoch time
+     * @param _endTime End time of the voting period in Unix Epoch time
+     * @param _checkpointId Index of the checkpoint to use for token balances
+     * @return bool success
+     */
+    function createCustomBallot(uint256 _startTime, uint256 _endTime, uint256 _checkpointId) public onlyOwner returns (bool) {
+        require(_endTime >= _startTime);
```
adamdossa: Should be > rather than >= I think (can't think of any reason why endTime == startTime makes sense). Can you add a require reason string (i.e. "start time is after end time).

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * @notice Allows the token issuer to create a ballot
+     * @param _duration The duration of the voting period in seconds
+     * @return bool success
+     */
+    function createBallot(uint256 _duration) public onlyOwner returns (bool) {
+        uint256 ballotId = ballots.length;
+        uint256 checkpointId = ISecurityToken(securityToken).createCheckpoint();
```
adamdossa, line 94: Would it be cleaner for createBallot to call createCustomBallot rather than copy / paste code?

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+     * @param _checkpointId Index of the checkpoint to use for token balances
+     * @return bool success
+     */
+    function createCustomBallot(uint256 _startTime, uint256 _endTime, uint256 _checkpointId) public onlyOwner returns (bool) {
+        require(_endTime >= _startTime);
+        uint256 ballotId = ballots.length;
+        uint256 supplyAtCheckpoint = ISecurityToken(securityToken).totalSupplyAt(_checkpointId);
+        ballots.push(Ballot(_checkpointId,supplyAtCheckpoint,_startTime,_endTime,0,0,0, true));
+        emit BallotCreated(_startTime, _endTime, ballotId, _checkpointId);
+        return true;
```
adamdossa: No need to return true on these types of functions (i.e. functions where the only possible return value is true)

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * @notice Allows the token issuer to set the active stats of a ballot
+     * @param _ballotId The index of the target ballot
+     * @param _isActive The bool value of the active stats of the ballot
+     * @return bool success
+     */
+    function setActiveStatsBallot(uint256 _ballotId, bool _isActive) public onlyOwner returns (bool) {
+        require(now > ballots[_ballotId].startTime && now < ballots[_ballotId].endTime, "Voting period is not active.");
```
maxsam4: I think we should allow this function to be called before voting period has started.

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+        ballots[_ballotId].isActive = _isActive;
+        emit BallotDeactivated(_ballotId, _isActive);
+        return true;
+    }
+
+    // /**
+    //  * @notice Init function i.e generalise function to maintain the structure of the module contract
+    //  * @return bytes4
+    //  */
```
maxsam4: Remove //

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @notice Allows the token issuer to set the active stats of a ballot
+     * @param _ballotId The index of the target ballot
+     * @param _isActive The bool value of the active stats of the ballot
+     * @return bool success
+     */
+    function setActiveStatsBallot(uint256 _ballotId, bool _isActive) public onlyOwner returns (bool) {
+        require(now > ballots[_ballotId].startTime && now < ballots[_ballotId].endTime, "Voting period is not active.");
+        ballots[_ballotId].isActive = _isActive;
+        emit BallotDeactivated(_ballotId, _isActive);
```
adamdossa: Probably event should be called something like BallotActiveStateChanged as it may be activating or deactivating the ballot.

adamdossa: Should also check that require(ballots[_ballotId].isActive != _isActive, "Active state unchanged")

contracts/modules/Checkpoint/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,148 @@
...
+        return true;
```
adamdossa: No need to return true on these types of functions

adamdossa (changes requested):

comeonbuddy:
@adamdossa  @maxsam4 fixed all comments.

adamdossa (approved):

PR #332, opened 2018-10-11
Renamed VRTM module

maxsam4:
Renamed `VolumeRestrictionTransferManager` to `LockupVolumeRestrictionTransferManager`.

Note
Name in factory has not been changed as `LockupVolumeRestrictionTransferManager` does not fit in `bytes32`


satyamakgec:
Couldn't we use the `TM` as the suffix for TransferManager that will reduce the contract name and also open the space for writing the clear name for the TM?
so it will be like `LockupVolumeRestrictionTM`, `PercentageTM`, `ManualApprovalTM`, `SingleTradeVolumeRestricitonTM` etc.
using this we can easily return the name as a `bytes32` datatype. 
@pabloruiz55 WDYT?

pabloruiz55:
@satyamakgec My concern is that introduces some discrepancies with older versions. 
Let's start by doing that with newer TM modules only for now. So, @maxsam4 for this one and single trade let's rename them to xxxTM

satyamakgec (approved):

PR #336, opened 2018-10-11
Revert "Voting module"

pabloruiz55:
Reverts PolymathNetwork/polymath-core#329


pabloruiz55:
Reverting this merge so we can complete documentation, improve tests and build the CLI before merging it. Also, so it's actually merged to 2.x and not to 1.5.0/2.0.0

PR #340, opened 2018-10-12
Migrating to yarn from npm

maxsam4:

Migrating to yarn from npm

A mixture of yarn and npm is used

### What is the new behavior?
Strictly yarn is used for adding/removing packages. I have removed package-lock.json and updated yarn.lock.

No.

### Any Other information:
npm isn't working for new users. Hence we have decided to move to yarn instead.
If you prefer to use npm then you can continue using npm but please don't push the package-lock.json to the repo. 
If you want to add a new node_module, then please use `yarn add module` instead of `npm install module` and push your yarn.lock only if it's clean.


satyamakgec:
@maxsam4 Please add `package-lock.json` in `.gitignore`. So if anyone wants to use it then they can on the local but it doesn't affect on the remote.

pabloruiz55 (approved):

PR #342, opened 2018-10-16
Increase coverage

satyamakgec:
Fixes during increasing the tests

- Wrong mapping used in the addManualBlocking() of MATM
- Return statement is missing from the implementation() of OwnerUpgradeability.sol
- change the `withdrawPoly` function with `withdrawERC20`.


contracts/SecurityTokenRegistry.sol
```diff
@@ -669,6 +669,7 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
     * @param _patch Patch version of the proxy
     */
     function setProtocolVersion(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) external onlyOwner {
+        require(_STFactoryAddress != address(0));
```
maxsam4: Missing reason string

contracts/tokens/SecurityToken.sol
```diff
@@ -341,13 +341,16 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
     * @dev Owner can transfer POLY to the ST which will be used to pay for modules that require a POLY fee.
+    * @param _tokenContract Address of the ERC20Basic compliance token
     * @param _value amount of POLY to withdraw
     */
-    function withdrawPoly(uint256 _value) external onlyOwner {
-        require(ERC20(polyToken).transfer(owner, _value), "Insufficient balance");
+    function withdrawERC20(address _tokenContract, uint256 _value) external onlyOwner {
+        require(_tokenContract != address(0));
+        IERC20 token = IERC20(_tokenContract);
+        require(token.transfer(owner, _value));
```
maxsam4, line 353: Missing reason strings

test/k_module_registry.js
```diff
@@ -133,6 +138,57 @@ contract("ModuleRegistry", accounts => {
         `);
     });
 
+    describe("Test the initialize the function", async () => {
+        it("Should successfully update the implementation address -- fail because polymathRegistry address is 0x", async () => {
+            let bytesProxy = encodeProxyCall(MRProxyParameters, [
+                "0x0000000000000000000000000000000000000000",
```
maxsam4: For cleaner code, maybe let's define `const nullAddress = "0x0000000000000000000000000000000000000000"` and use it everywhere. It is used a lot in all the test cases.

contracts/mocks/MockBurnFactory.sol
```diff
@@ -0,0 +1,101 @@
...
+
+import "./MockRedemptionManager.sol";
+import "../modules/ModuleFactory.sol";
+import "../libraries/Util.sol";
+
+/**
+ * @title Mock Contract Not fit for production environment
+ */
+
+contract MockBurnFactory is ModuleFactory {
```
maxsam4: I think we should inherit the main contract in the mock contract and overwrite the functions we need to. This will help the reader understand what exactly we are mocking and why. 
`contract MockBurnFactory is ModuleFactory` 
can be 
`contract MockBurnFactory is BurnFactory`

Same goes for all mock contracts.

maxsam4 (changes requested):

contracts/tokens/SecurityToken.sol
satyamakgec: It is intentional to reduce the size of the contract

maxsam4: I think we should add at least 1 character reason so that we can know if our require statement is reverting or if it's something generic.
I'll have a deep look at the security token contract and to see if I can save some bits at some other places.

maxsam4 (approved):

PR #343, opened 2018-10-16
Windows support

maxsam4:
`npm run wintest` -> Run tests on windows
`npm run wincov` -> Run coverage on windows


VictorVicente:
I'll try it in an hour, looks good.
What do you think about using a package like run-script-os (https://www.npmjs.com/package/run-script-os) or similar, to have an unified test command?

maxsam4:
> What do you think about using a package like run-script-os to have an unified test command?

I am neutral on the issue. On one hand, we get unified command and on the other, an additional dependency. I'll see if I can make our script unified without the need of an additional package.

maxsam4:
I was able to make a unified script but it needs to be placed in root directory of the repo as path definition are different on windows. I would prefer using the node module or non-unified commands over placing a script in the root of the repo.

PR #344, opened 2018-10-17
Dividend delegate

maxsam4:
Right now there's only a "DISTRIBUTE" permission for pushing dividends.
The Issuer should be able to delegate the creation of checkpoints / dividends and the management of tax withholding / exclusions to delegates.

I have added the basic functionality and test case. I would ideally like to add _delegate_ test case for every function before merging. Putting this on hold for now in favor of a different task.

Meanwhile, This can be reviewed to verify that it does what the intention of the task is.

UPDATE: Added remaining test cases. Can be reviewed now.


satyamakgec:
The code is good, But we can divide the permission at more granular level. something like that 
**DISTRIBUTE** - Use to distribute the dividends  functions - `pushDividendPaymentToAddresses()` , `pushDividendPayment()`.
**MANAGE** - Use to set the values related to investors (Kind of management) `setDefaultExcluded()`, `setWithholding()`, `setWithholdingFixed()`.
**CREATE** - use to create the dividends - `createDividend()`, `createDividendWithCheckpoint()`, `createDividendWithExclusions()`, `createDividendWithCheckpointAndExclusions()`.
**WITHDRAW** - used for extraction of crypto from smart contract `reclaimDividend()`, `withdrawWithholding()` - This one we can be used by changing` msg.sender` to `ISecurityToken(SecurityToken).owner()`.  I am thinking that is useful because it will kind of regular process of providing the dividends so may be Issuer also want to delegate the permission of accessing these function to someone whom he/she can trust. 
@pabloruiz55 WDYT?

maxsam4:
@satyamakgec 
please refer https://app.asana.com/0/792502492975207/867290984849674

pabloruiz55:
@satyamakgec yeah, I didn't want to overcomplicate it. Let's do just one permission level and if needed in the future we can release an update with better permissions.

VictorVicente (approved):

PR #345, opened 2018-10-17
Multiple controllers

maxsam4:

Added support for multiple controllers

Only a single controller/operator is allowed

### What is the new behavior?
Multiple controllers/operators are allowed

- `controllerDisabled` variable renamed to `forceTransferDisabled`. `DisableController` event renamed to `DisableForceTransfer`. To disable controllers, use `setController` function.
- `controller` variable is now a mapping instead of a single address.
- `createCheckpoint` can now also be called by controllers
- `setController` (function and event) now takes a second variable (boolean) which defines if the controller is being activated or deactivated. (true -> add controller | false -> remove controller)

### Any Other information:
Gas cost for generating ST rose from 7.54m to 7.56m.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 [__1.5.0__](https://www.npmjs.com/package/polymath-core?activeTab=readme) __15-08-18__
 
 ## Added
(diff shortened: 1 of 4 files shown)


satyamakgec (approved):
Looks good to me! @pabloruiz55 please take a second look

pabloruiz55:
Closing and dismissing this PR according to the conversation we had today.

PR #347, opened 2018-10-17
change in the permission doc

satyamakgec:
Change in the permission doc related to new Modules.
After merging the checkpoint PR. I will update the perm for the checkpoint module


satyamakgec:
@pabloruiz55 now you can merge this

pabloruiz55 (approved):

PR #356, opened 2018-10-20
[Fix] Duplicate Address in exclusions[]

maxsam4:

Duplicate and zero address are no longer allowed in exclusions list when creating a dividend with exclusion or setting default exclusions.

### DIFF
--- a/contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
+++ b/contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
@@ -124,9 +124,7 @@ contract EtherDividendCheckpoint is DividendCheckpoint {
 
         for (uint256 j = 0; j < _excluded.length; j++) {
             require (_excluded[j] != address(0), "Invalid address");
-            for (uint256 i = j + 1; i < _excluded.length; i++) {
-                require (_excluded[j] != _excluded[i], "Duplicate exclude address");
-            }
+            require(!dividends[dividendIndex].dividendExcluded[_excluded[j]], "duped exclude address");
             excludedSupply = excludedSupply.add(ISecurityToken(securityToken).balanceOfAt(_excluded[j], _checkpointId));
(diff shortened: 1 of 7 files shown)


maxsam4:
Seems like the online GitHub merge resolver added whitespace changes to the whole contracts.

maxsam4:
Shortened a few strings so that ERC20DividendFactory can be deployed within the limit.

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -121,11 +121,16 @@ contract EtherDividendCheckpoint is DividendCheckpoint {
             _name
           )
         );
-        for (uint256 i = 0; i < _excluded.length; i++) {
-            excludedSupply = excludedSupply.add(ISecurityToken(securityToken).balanceOfAt(_excluded[i], _checkpointId));            
-            dividends[dividends.length - 1].dividendExcluded[_excluded[i]] = true;
+
+        for (uint256 j = 0; j < _excluded.length; j++) {
+            require (_excluded[j] != address(0), "Invalid address");
+            for (uint256 i = j + 1; i < _excluded.length; i++) {
```
adamdossa: Instead of the two loops, could we just have:
`require(!dividends[dividendIndex].dividendExcluded[_excluded[i]], "duped address")`
?

adamdossa: (same for ERC20 & EtherDividends)

adamdossa (changes requested):

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
maxsam4: Reading will storage will cost ~250 gas so this will become effective only is the list size is greater than ~20.
I'll still make the change because it lowers the code complexity.
Thanks for pointing it out.

adamdossa (approved):

PR #368, opened 2018-10-22
Optimized verifyTransfer

maxsam4:
Minor Optimization

### DIFF
--- a/contracts/tokens/SecurityToken.sol
+++ b/contracts/tokens/SecurityToken.sol
@@ -547,7 +547,6 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
             }
             bool isInvalid = false;
             bool isValid = false;
-            bool isForceValid = false;
             bool unarchived = false;
             address module;
             for (uint8 i = 0; i < modules[TRANSFER_KEY].length; i++) {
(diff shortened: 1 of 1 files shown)


pabloruiz55:
@adamdossa please review and merge if ok.

contracts/tokens/SecurityToken.sol
```diff
@@ -540,34 +540,28 @@ contract SecurityToken is StandardToken, DetailedERC20, ReentrancyGuard, Registr
...
-                    }
-                    if (valid == ITransferManager.Result.VALID) {
+                    } else if (valid == ITransferManager.Result.VALID) {
                         isValid = true;
-                    }
-                    if (valid == ITransferManager.Result.FORCE_VALID) {
-                        isForceValid = true;
+                    } else if (valid == ITransferManager.Result.FORCE_VALID) {
+                        return true;
                     }
```
adamdossa, line 569: @maxsam4 I don't think we can do this as we need to guarantee that all TMs see all transfers (e.g. in case they are maintaining internal state related to the transfer). If _isTransfer == false we could bypass, but I think it is cleaner to just ensure that all TMs see all transfers rather than breaking out of the loop by returning `true` in the FORCE_VALID case.

adamdossa (changes requested):

contracts/tokens/SecurityToken.sol
maxsam4: Yeah, that makes sense. Gonna push a fix in a bit. So basically, the only optimizations left are using `else if` rather than `if` and removal of length check.

adamdossa (approved):

PR #380, opened 2018-10-25
ModuleFactory var visibility changed to internal

kostind:

ModileFactory variables visibility has been changed to internal

### What is the new behavior?

### DIFF
--- a/contracts/modules/ModuleFactory.sol
+++ b/contracts/modules/ModuleFactory.sol
@@ -15,11 +15,11 @@ contract ModuleFactory is IModuleFactory, Ownable {
     uint256 public usageCost;
     uint256 public monthlySubscriptionCost;
 
-    uint256 internal setupCost;
-    string internal description;
-    string internal version;
-    bytes32 internal name;
-    string internal title;
+    uint256 public setupCost;
+    string public description;
+    string public version;
+    bytes32 public name;
+    string public title;
 
     // @notice Allow only two variables to be stored
     // 1. lowerBound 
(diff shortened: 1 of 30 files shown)


contracts/modules/ModuleFactory.sol
```diff
@@ -12,13 +12,14 @@ import "../libraries/VersionUtils.sol";
 contract ModuleFactory is IModuleFactory, Ownable {
 
     IERC20 public polyToken;
-    uint256 public setupCost;
     uint256 public usageCost;
     uint256 public monthlySubscriptionCost;
```
satyamakgec, line 16: I don't understand How it will increase the gas cost? could you share please share some errors with me that you got during the changes or some stats (How much gas increase for the deployment). 
If we are marking all other variables `internal` then for consistency make `usageCost` and `monthlySubscriptionCost` internal too and create the corresponding getters for them (This will also increase slight a bit of gas cost of deployment).

satyamakgec (dismissed):

contracts/modules/ModuleFactory.sol
kostind: I've got the following error for "npm run wintest" after removing getters:
Error: The contract code couldn't be stored, please check your gas amount.
Fixed by increased gas limit for development network from 7900000 to 8900000
Now I have another one
Error: Exceeds block gas limit

You can see changes in a separate branch - Remove-redundant-getters

pabloruiz55 (approved):

PR #387, opened 2018-10-26
USDTieredSTO description

satyamakgec:


F-OBrien:
Description grammatical error needs correction. Should be "Tokens will be sold" or "Tokens are sold" not " tokens will sold"

pabloruiz55 (approved):

PR #390, opened 2018-10-26
[CLI] Remove defaultGasPrice and 'from' from SendTransaction

shuffledex:


CLI/commands/common/constants.js
```diff
@@ -0,0 +1,35 @@
+module.exports = Object.freeze({
+    NETWORK: undefined,
```
VictorVicente: I think NETWORK shouldn't be a constant. It should be a global variable, like 'Issuer' or 'defaultGasPrice'

CLI/polymath-cli.js
```diff
@@ -140,20 +141,22 @@ program
...
+    await permission_manager.executeApp();
   });
 
 program.parse(process.argv);
 
 if (typeof program.commands.length == 0) {
   console.error('No command given!');
   process.exit(1);
+} else {
+  global.constants.NETWORK = program.remoteNode
```
VictorVicente: Instead of this, maybe we could call to global.Initialize(program.remoteNode) here. This way, we could remove initialize command from everywhere and call it only once.

CLI/commands/common/common_functions.js
```diff
@@ -25,6 +26,38 @@ async function checkPermission(contractName, functionName, contractRegistry) {
   }
 };
 
+function getFinalOptions(options) {
```
VictorVicente, line 28: It should check if 'options' is undefined. This way we don't need to send '{}' on every sendTransaction and it will be totally optional.

CLI/commands/common/common_functions.js
```diff
@@ -38,78 +71,52 @@ module.exports = {
...
+  sendTransaction: async function (action, options) {
 
-    let block = await web3.eth.getBlock("latest");
-    let networkGasLimit = block.gasLimit;
+    await checkPermissions(action);
 
-    let gas = Math.round(factor * (await action.estimateGas({ from: from.address, value: value})));
-    if (gas > networkGasLimit) gas = networkGasLimit;
+    options = getFinalOptions(options);
+    let gas = await getGas(options, action);
```
VictorVicente: Replace 'gas' for 'gasLimit' and 'getGas' for 'getGasLimit'

CLI/commands/common/common_functions.js
```diff
@@ -25,6 +26,38 @@ async function checkPermission(contractName, functionName, contractRegistry) {
   }
 };
 
+function getFinalOptions(options) {
+  const defaultOptions = {
```
VictorVicente, line 32: We could move this to constants.js

VictorVicente (changes requested):

CLI/commands/common/constants.js
shuffledex: I didn't know in what I was thinking....

CLI/polymath-cli.js
shuffledex: Good idea! I'll figure out how could I implement it.

CLI/commands/common/common_functions.js
shuffledex: I understand, but the idea behind the explicit '{}' is remember that you could overwrite any params regarding tx.

CLI/commands/common/common_functions.js
VictorVicente: This is missing, isn't it?

CLI/polymath-cli.js
VictorVicente: Is it possible to replace all the occurrences of 'Initialize()' for a single line here? I don't remember how Commander works, but it could work.

shuffledex: it's not possible. We haven't the ability to run an action before the command catcher. I had to modify every action to perform the initialize method outside the module controller.

CLI/commands/common/common_functions.js
shuffledex: I can't move it 'cos the constants were defined before Issuer and defaultGasPrice.

VictorVicente: Makes sense

VictorVicente:
@shuffledex I found that scripts commands (whitelist.js, multi_mint.js, accredit.js and changeNonAccreditedLimit.js) fails if they are run from another command. This is because they don't find web3 instance.

VictorVicente:
gbl.constants.NETWORK doesn't exists anymore. 
It should be replaced by a global variable like Issuer or defaultGasPrice at Initialize(). 
@shuffledex check it and fix it on ST20Generator.js and strMigrator.js

CLI/commands/dividends_manager.js
```diff
@@ -362,7 +344,7 @@ async function createDividends(name, dividend, checkpointId) {
         createDividendAction = currentDividendsModule.methods.createDividendWithExclusions(maturityTime, expiryTime, excluded, web3.utils.toHex(name));
       }
     }
-    let receipt = await common.sendTransaction(Issuer, createDividendAction, defaultGasPrice, web3.utils.toWei(dividend));
+    let receipt = await common.sendTransaction(createDividendAction, {gasPrice: web3.utils.toWei(dividend)});
```
VictorVicente: This should be 'value' instead of 'gasPrice'

VictorVicente (changes requested):
Only one little thing to fix

VictorVicente (approved):

PR #391, opened 2018-10-27
Single lockup only

glitch003:
This is the final fix per @satyamakgec request to make the VolumeRestrictionTransferModule only handle a single lockup.


maxsam4:
Continuing the development on https://github.com/PolymathNetwork/polymath-core/pull/409

PR #396, opened 2018-11-07
[CLI] Show delegates and their permission

shuffledex:


CLI/commands/permission_manager.js
```diff
@@ -239,6 +244,46 @@ async function getModulesWithPermissions() {
...
+  }
+  return result
+}
+
+function renderTable(permisions, address) {
+  let result = ``;
+  if (permisions[address] != undefined) {
+    for (const obj of permisions[address]) {
+      result += `
+      ${obj.module} -> ${obj.permission}`;
```
VictorVicente: Instead of repeat each module for every permission granted, it should group all perms for each module in one line.

VictorVicente (changes requested):
Looks good, only one change requested.

CLI/commands/permission_manager.js
shuffledex: makes sense!

VictorVicente:
Please, add the module address too. It will be necessary when there are more than one module with the same name.

VictorVicente (approved):

PR #397, opened 2018-11-08
USD Tiered STO changes

maxsam4:
changelog
1) Calculate the tokens minted so it's a round number and send back the change to the investor.
(Instead of attempting to purchase a non-round amount of tokens and have it revert because of granularity)
2) Adding an option for buyers to set maximum buy price (Audit suggestion).
3) Adding getter function for dApp.


maxsam4:
Test cases added, comments cleaned up
TODO: update changelog, look into getSTODetails, revert to classic factory.

maxsam4:
It still did not fit in the factory so we are gonna stick with proxy deployer for now.
Added getSTODetails to it and updated changelog.

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -197,33 +192,26 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
+        _modifyTimes(_startTime, _endTime);
+        _modifyTiers(_ratePerTier, _ratePerTierDiscountPoly, _tokensPerTierTotal, _tokensPerTierDiscountPoly);
+        // NB - _setFundRaiseType must come before modifyAddresses
+        _setFundRaiseType(_fundRaiseTypes);
+        _modifyAddresses(_wallet, _reserveWallet, _usdToken);
+        _modifyLimits(_nonAccreditedLimitUSD, _minimumInvestmentUSD);
     }
 
-    function modifyFunding(FundRaiseType[] _fundRaiseTypes) public onlyFactoryOrOwner {
+    function modifyFunding(FundRaiseType[] _fundRaiseTypes) external {
```
satyamakgec: Why its permission gets removed? Now anybody can change the fundRaiseType

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -197,33 +192,26 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
-        for (uint8 i = 0; i < _fundRaiseTypes.length; i++) {
-            mintedPerTier[uint8(_fundRaiseTypes[i])] = new uint256[](length);
-        }
     }
 
     function modifyLimits(
         uint256 _nonAccreditedLimitUSD,
         uint256 _minimumInvestmentUSD
-    ) public onlyFactoryOrOwner {
+    ) external onlyFactoryOrOwner {
```
satyamakgec: use the internal version of modifyLimits() ~ _modifyLimits(). and use the OnlyOwner permission

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -234,49 +222,80 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
         uint256[] _ratePerTierDiscountPoly,
         uint256[] _tokensPerTierTotal,
         uint256[] _tokensPerTierDiscountPoly
-    ) public onlyFactoryOrOwner {
+    ) external onlyFactoryOrOwner {
```
satyamakgec: use onlyOwner modifier here

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -234,49 +222,80 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
-        emit SetTiers(_ratePerTier, _ratePerTierDiscountPoly, _tokensPerTierTotal, _tokensPerTierDiscountPoly);
+        require(now < startTime, "STO already started");
+        _modifyTiers(_ratePerTier, _ratePerTierDiscountPoly, _tokensPerTierTotal, _tokensPerTierDiscountPoly);
     }
 
     function modifyTimes(
         uint256 _startTime,
         uint256 _endTime
-    ) public onlyFactoryOrOwner {
+    ) external onlyFactoryOrOwner {
```
satyamakgec: use onlyOwner() modifier

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -234,49 +222,80 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
+        _modifyTimes(_startTime, _endTime);
+    }
+
+    function modifyAddresses(
+        address _wallet,
+        address _reserveWallet,
+        address _usdToken
+    ) external onlyFactoryOrOwner {
```
satyamakgec: use onlyOwner() modifer

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -234,49 +222,80 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
+        uint256[] _tokensPerTierTotal,
+        uint256[] _tokensPerTierDiscountPoly
+    ) internal {
+        require(_tokensPerTierTotal.length > 0, "No tiers provided");
+        require(_ratePerTier.length == _tokensPerTierTotal.length &&
+            _ratePerTierDiscountPoly.length == _tokensPerTierTotal.length &&
+            _tokensPerTierDiscountPoly.length == _tokensPerTierTotal.length, 
+            "Tier data length mismatch"
+        );
+        delete tiers;
```
satyamakgec, line 297: This function is allowing to change the no. of tiers. If STO has 5 tiers and If Issuer wants only to change the 3 tiers data (Allowed by the function no restriction that checked the length of arrays should be equal to no. of tiers) then this function will first delete all the 5 tiers data and then push the 3 new tiers data.
Possible simple solution:
add this `require(_ratePerTier.length == tiers.length);`  in modifyTiers() function.

So using the above require statement issuer always get restricted to change the all tiers data at once. AFAICT this was not a problem in the previous implementation.

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -358,16 +377,32 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
-        buyWithETH(msg.sender);
+        buyWithETHRateLimited(msg.sender, 0);
+    }
+
+    // For backwards compatibility
+    function buyWithETH(address _beneficiary) public payable {
+        buyWithETHRateLimited(_beneficiary, 0);
+    }
+
+    function buyWithPOLY(address _beneficiary, uint256 _investedPOLY) public {
```
satyamakgec: add validPOLY() modifier

satyamakgec: prefer to call buyWithPOLYRateLimited() instead of _buyWithTokens()

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -358,16 +377,32 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
+        _buyWithTokens(_beneficiary, _investedPOLY, FundRaiseType.POLY, 0);
+    }
+
+    function buyWithUSD(address _beneficiary, uint256 _investedDAI) public validDAI {
+        _buyWithTokens(_beneficiary, _investedDAI, FundRaiseType.DAI, 0);
```
satyamakgec: prefer to call buyWithPOLYRateLimited() instead of _buyWithTokens() for consistency in all functions

satyamakgec (changes requested):
not fully reviewed yet (WIP) but you can correct those if you want @maxsam4

contracts/modules/STO/USDTieredSTO.sol
maxsam4: Thanks for catching this. I must have changed it by mistake. Will add onlyOwner.

contracts/modules/STO/USDTieredSTO.sol
maxsam4: I kept the permission same as earlier but you're right, we don't need factory permission. Will change in all.

contracts/modules/STO/USDTieredSTO.sol
maxsam4: This is not a problem IMO, this is as I intended it to be. I wanted issuer to be able to change the number of tiers. The dApp can be programmed to send existing data to the function along with changed data.

maxsam4: Also, the same thing was happening in the last implementation as well. You were not allowed to edit specific tiers. You needed to feed all tiers to the function even if you want to edit only one. The only difference is that I have explicitly added `delete` while the last implementation just overwrites on storage. 

If needed, I can add a separate function that will edit one tier at a time. Although I believe that dApp should be able to handle that.

satyamakgec: Yes, In last implementation you can't change the specific tier data but if I want to change the data of the first three tiers then you can change also that will also not effect on the current no. of tiers.

Yes dApp can be programmed but we also need to take care when someone uses our smart contract directly (without any dApp).

maxsam4: No, you can't change data of only first 3 tiers in the last implementation. The code replaces the storage array with one passed in the function. If the input had only 3 tiers data, Storage will now have only 3 tiers data.

I'll add the function to edit a single tier. Although I highly doubt that anyone will use smart contracts directly for a function available in the dApp.

maxsam4: There isn't even an event to show single tier change. I can populate the list in a memory array for the event but I just can't justify it. We can't even edit other parameters like startTime / endTime separately with current code. The file size is already so big that we are forced to use proxy deployment. I don't think we should be adding this feature.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 [__1.5.0__](https://www.npmjs.com/package/polymath-core?activeTab=readme) __15-08-18__
 
 ## Added
+* Added `getSTODetails` to USDTSTO.
+* Added an Array of Tiers that will hold data about every tier in USDTSTO.
+* Added `buyWithETHRateLimited`, `buyWithPOLYRateLimited` and `buyWithUSDRateLimited` to USDTSTO.
```
satyamakgec, line 12: create the 2.x.x release and add these changes into that.

CHANGELOG.md
```diff
@@ -42,6 +45,7 @@ All notable changes to this project will be documented in this file.
 * Change the function name of `withdraPoly` to `withdrawERC20` and make the function generalize to extract tokens from the ST contract. parmeters are contract address and the value need to extract from the securityToken.          
 
 ## Removed
+* Individual mappings for tier data removed in UDSTSTO.
```
satyamakgec: create the 2.x.x release and add these changes into that.

contracts/modules/STO/USDTieredSTO.sol
satyamakgec: okay cool, could you please add the developer comment about the behavior of the function.

satyamakgec (changes requested):

satyamakgec:
@maxsam4 could you please elaborate why you are using the `_minTokens` check in the contract. as far I understand it is only used to check the minimum demand of tokens for a buy call.

maxsam4:
@satyamakgec Yes, you are right. It is one of the Audit suggestions that a user must be able to set the minimum amount of tokens bought in the STO or the min conversion rate and max token rate. This is because the rate can be different at the time transaction is created and mined. I have kept the old functions intact so if a user does not want to limit the rate, they can choose to do so. As an investor, I don't want to buy tokens at a price much higher than what I signed up for. This can happen in case where POLY discount token limit is reached by some other tx before our tx is mined, tier changes before our tx is mined, eth suddenly crashes etc.

CHANGELOG.md
maxsam4: Will do it along with missing code documentation tomorrow.

satyamakgec:
I am done with the reviewing part @adamdossa please take a final look.

adamdossa (approved):

adamdossa:
@maxsam4 I reviewed and approved. Just one question - what was the reason to push the `gotoNextTier` logic down into the function rather than the previous logic?

maxsam4:
> @maxsam4 I reviewed and approved. Just one question - what was the reason to push the `gotoNextTier` logic down into the function rather than the previous logic?

With the new logic, only granular units of tokens are purchased which means not all the InvestedUSD may be consumed. We don't need to go to the next tier in this scenario.

For example, If I have an Indivisible token which costs 10 a pop and an Investor invests 15 USD then the first tier will buy him 1 token and return from the function. Earlier logic would have seen pending 5 USD and called the next tier. In this case, we don't need to call the next tier. We just need to refund the 5USD back to the investor.

adamdossa:
Got it thanks - that makes sense.

satyamakgec (approved):

maxsam4:
@VictorVicente I've added view functions that were required for CLI and dApp.

@satyamakgec @adamdossa I've made some structural changes in `_buyTokens` and its internal functions to reduce code redundancy. Please review again.

I'll add test cases for the new `view` functions in a bit.

F-OBrien:
typo at line 487 of investor_portal.js "You are going to spen**t**" should be "you are going to spen**d**"

maxsam4:
> typo at line 487 of investor_portal.js "You are going to spen**t**" should be "you are going to spen**d**"

Thanks for pointing that out @FOBCrypto .
Would you like to use the new GitHub beta suggestion feature to suggest that change? That way the commit will be linked to your GitHub profile and you'll officially become a contributor :).

CLI/commands/investor_portal.js
```diff
@@ -481,6 +482,14 @@ async function investUsdTieredSTO(currency, amount) {
 
     let costWei = web3.utils.toWei(cost.toString());
 
+    let tokensToBuy = await currentSTO.methods.buyTokensView(User.address, costWei, gbl.constants.FUND_RAISE_TYPES[raiseType]).call();
+    let minTokenToBuy = tokensToBuy.tokensMinted;
+    console.log(chalk.yellow(`You are going to spent ${web3.utils.fromWei(tokensToBuy.spentValue)} ${raiseType} (${web3.utils.fromWei(tokensToBuy.spentUSD)} USD) to buy ${web3.utils.fromWei(minTokenToBuy)} ${STSymbol} approx.`));
```
F-OBrien:
```suggestion
    console.log(chalk.yellow(`You are going to spend ${web3.utils.fromWei(tokensToBuy.spentValue)} ${raiseType} (${web3.utils.fromWei(tokensToBuy.spentUSD)} USD) to buy ${web3.utils.fromWei(minTokenToBuy)} ${STSymbol} approx.`));
```

PR #405, opened 2018-11-10
getTickersOwner fix

satyamakgec:


adamdossa (approved):

test/n_security_token_registry.js
```diff
@@ -993,7 +993,8 @@ contract("SecurityTokenRegistry", accounts => {
 
         it("Should get the tickers by owner", async () => {
             let tickersList = await I_STRProxied.getTickersByOwner.call(token_owner);
-            assert.equal(tickersList.length, 4);
+            console.log(tickersList);
+            assert.equal(tickersList.length, 3);
```
maxsam4:
```suggestion
            assert.equal(tickersList.length, 4);
```

maxsam4: 4 Tickers(6 but 2 are being reverted due to snapshot restore) are being registered by the token_owner in the tests. Not sure why it was changed. Fix looks good otherwise. Nice quick work @satyamakgec!

satyamakgec: hmm in local test it was reverting, because 4th ticker was null because of that bug so it should be 3 tickers. let me run it again to confirm.

satyamakgec: on local if I run without the fix then it give me this 
 ['0x4445540000000000000000000000000000000000000000000000000000000000',
  '0x4554480000000000000000000000000000000000000000000000000000000000',
  '0x504f4c5900000000000000000000000000000000000000000000000000000000',
  '0x0000000000000000000000000000000000000000000000000000000000000000' ]
so 4th is null that should not be added in the output. so I run the fixed code then It always return the 3 ticker 
 ['0x4445540000000000000000000000000000000000000000000000000000000000',
  '0x4554480000000000000000000000000000000000000000000000000000000000',
  '0x504f4c5900000000000000000000000000000000000000000000000000000000' ]
I don't understand why it is not reflecting on the CI. 
Note - I run individual test file not the full test suite

maxsam4: Contracts are reset by truffle on every test file so running individually should not matter.
In the test code, I can see 4 tickers being registered. (lines 387,409,549,575)

Issue #389, opened 2018-10-26
Audit Issue - LockupVolumeRestrictionTM.sol


adamdossa:
The following scenario still breaks the intended behavior of LockupVolumeRestrictionTM, "Lock 3 wei, wait for two to be unlocked but transfer only one, then wait for the final one to unlock and try to transfer the remaining two, this should be possible but isn't". Though the contract is currently not meant for production, this test case should be documented for future reference.

adamdossa:
@satyamakgec Trying to understand this bug a bit more...

AFAICT, in this scenario:

`uint elapsedPeriods = (now.sub(lockUps[userAddress].startTime)).div(lockUps[userAddress].releaseFrequencySeconds);`
=> `elapsedPeriods == 3`
`uint allowedAmount = amountPerPeriod.mul(elapsedPeriods).sub(lockUps[userAddress].alreadyWithdrawn);`
=> `allowedAmount == 3 - 1 == 2`

So you would be able to withdraw the 2 WEI. Is there something I'm missing?

satyamakgec:
I don't think so, It is working as intended

PR #409, opened 2018-11-13
Reinvent lockup

satyamakgec:
- [X] Test suite


satyamakgec:
Please do check with the issue #389 As per the test it is not present in the current code

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +282,104 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-        ) internal view returns (Result) {
-        // the amount the user wants to withdraw is greater than their allowed amounts according to the lockups.  however, if the user has like, 10 tokens, but only 4 are locked up, we should let the transfer go through for those 6 that aren't locked up
-        uint currentUserBalance = ISecurityToken(securityToken).balanceOf(userAddress);
-        uint stillLockedAmount = totalSum.sub(alreadyWithdrawnSum);
-        if (currentUserBalance >= stillLockedAmount && amount <= currentUserBalance.sub(stillLockedAmount)) {
-            // the user has more tokens in their balance than are actually locked up.  they should be allowed to withdraw the difference
+        // Present balance of the user
+        uint256 currentBalance = ISecurityToken(securityToken).balanceOf(_userAddress);
+        if ((currentBalance.sub(_amount)) >= totalRemainingLockedAmount) {
             return Result.VALID;
```
maxsam4: I think we should return `Result.NA` instead of `Result.VALID`

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,218 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
     }
 
-
     /** @notice Used to verify the transfer transaction and prevent locked up tokens from being transferred
      * @param _from Address of the sender
      * @param _amount The amount of tokens to transfer
-     * @param _isTransfer Whether or not this is an actual transfer or just a test to see if the tokens would be transferrable
      */
-    function verifyTransfer(address  _from, address /* _to*/, uint256  _amount, bytes /* _data */, bool  _isTransfer) public returns(Result) {
+    function verifyTransfer(address  _from, address /* _to*/, uint256  _amount, bytes /* _data */, bool /*_isTransfer*/) public returns(Result) {
```
maxsam4: Can be declared `view`

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,218 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
      * @notice Lets the admin create a volume restriction lockup for a given address.
      * @param _userAddress Address of the user whose tokens should be locked up
+     * @param _lockupAmount Amount of tokens that need to lock.
+     * @param _startTime When this lockup starts (seconds)
      * @param _lockUpPeriodSeconds Total period of lockup (seconds)
      * @param _releaseFrequencySeconds How often to release a tranche of tokens (seconds)
-     * @param _startTime When this lockup starts (seconds)
-     * @param _totalAmount Total amount of locked up tokens
      */
     function addLockUp(
```
maxsam4: This can be broken into an internal function so that when calling `addLockUpMulti`, `withPerm(ADMIN)` is not checked again and again.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +282,104 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
+                // Calculate the no. of periods for a lockup 
+                uint256 noOfPeriods = (userLockup[_lockupIndex].lockUpPeriodSeconds).div(userLockup[_lockupIndex].releaseFrequencySeconds);
+                // Calculate the transaction time lies in which period
+                /*solium-disable-next-line security/no-block-members*/
+                uint256 elapsedPeriod = (now.sub(userLockup[_lockupIndex].startTime)).div(userLockup[_lockupIndex].releaseFrequencySeconds);
+                // Calculate the allowed unlocked amount per period
+                uint256 amountPerPeriod = (userLockup[_lockupIndex].lockupAmount).div(noOfPeriods);
+                // Find out the unlocked amount for a given lockup
+                uint256 unLockedAmount = elapsedPeriod.mul(amountPerPeriod);
+                return unLockedAmount;
```
maxsam4: Rounding off can cause tokens to be locked forever. Consider this example:
Granularity of token = 1
Total locked amount = 10
Number of periods = 6

This means, the amount per lock period = 10/6 = 1.67 = 1.
unlocked amount even after all periods are passed = 6*1 = 6
10 - 6 = 4 tokens are locked forever.
 
4 is a very small number as 1 token is 10**18 but it can cause amount mismatch and confusion as it will make transactions revert when user will try to transfer all of locked tokens.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +282,104 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-        uint totalPeriods = lockUpPeriodSeconds.div(releaseFrequencySeconds);
-        require(
-            totalAmount % totalPeriods == 0,
-            "The total amount being locked up must be evenly divisible by the number of total periods"
-        );
-
         // make sure the amount to be released per period is not too granular for the token
-        uint amountPerPeriod = totalAmount.div(totalPeriods);
+        uint256 totalPeriods = _lockUpPeriodSeconds.div(_releaseFrequencySeconds);
+        uint256 amountPerPeriod = _lockupAmount.div(totalPeriods);
```
maxsam4: To fix the issue I described above, maybe add `require(amountPerPeriod*totalPeriods == totalAmount);`

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,218 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
      * @param _userAddress Address of the user whose tokens are locked up
-     * @param _lockUpIndex The index of the LockUp to remove for the given userAddress
+     * @param _lockupIndex Index of the lockup need to be removed.
      */
-    function removeLockUp(address _userAddress, uint _lockUpIndex) public withPerm(ADMIN) {
-        LockUp[] storage userLockUps = lockUps[_userAddress];
-        require(_lockUpIndex < userLockUps.length, "Array out of bounds exception");
-
-        LockUp memory toRemove = userLockUps[_lockUpIndex];
+    function removeLockUp(address _userAddress, uint256 _lockupIndex) public withPerm(ADMIN) {
```
maxsam4: This can also be broken into internal and external functions.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,218 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-     * @param _lockUpIndex The index of the LockUp to edit for the given userAddress
+     * @param _lockupAmount Amount of tokens that needs to be locked
+     * @param _startTime When this lockup starts (seconds)
      * @param _lockUpPeriodSeconds Total period of lockup (seconds)
      * @param _releaseFrequencySeconds How often to release a tranche of tokens (seconds)
-     * @param _startTime When this lockup starts (seconds)
-     * @param _totalAmount Total amount of locked up tokens
+     * @param _lockupIndex Index of the lockup that needs to be modified.
      */
     function modifyLockUp(
```
maxsam4: This can also be broken down into internal and external functions.

maxsam4 (changes requested):
Consider adding a view `getLockedToken` function which will return sum of cirrently locked tokens of a user so that user can know how many tokens are still locked and how many can they transfer.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
satyamakgec: hmm I think I add a check for that in `_checkLockUpParams()`. I need to improve that check to solve this. I am not allowing to create a lockup with these figures( as a quick solution to reduce the complexity).

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
satyamakgec: yes that I said above

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
satyamakgec: should be complaint with` ITransferManager.sol`

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,171 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-
-        lockUps[_userAddress].push(LockUp(_lockUpPeriodSeconds, _releaseFrequencySeconds, startTime, _totalAmount, 0));
-
-        emit AddNewLockUp(
+        uint256 _lockupAmount,
+        uint256 _startTime,
+        uint256 _lockUpPeriodSeconds,
+        uint256 _releaseFrequencySeconds
+    ) 
+        public
```
maxsam4:
```suggestion
        external 
```

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,171 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-
-        // Get the lockup from the master list and edit it
-        lockUps[_userAddress][_lockUpIndex] = LockUp(
+        uint256 _lockupAmount,
+        uint256 _startTime,
+        uint256 _lockUpPeriodSeconds,
+        uint256 _releaseFrequencySeconds,
+        uint256 _lockupIndex
+    ) 
+        public
```
maxsam4:
```suggestion
        external 
```

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -61,166 +62,171 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-
-        emit RemoveLockUp(
-            _userAddress,
-            toRemove.lockUpPeriodSeconds,
-            toRemove.releaseFrequencySeconds,
-            toRemove.startTime,
-            toRemove.totalAmount,
-            _lockUpIndex
-        );
+    function removeLockUp(address _userAddress, uint256 _lockupIndex) public withPerm(ADMIN) {
```
maxsam4:
```suggestion
    function removeLockUp(address _userAddress, uint256 _lockupIndex) external withPerm(ADMIN) {
```

maxsam4:
TIL that references to parameters of an external function are stored in the stack.

satyamakgec:
@maxsam4 It is done with the review? If yes then @VictorVicente can use this same PR to create the CLI

maxsam4 (approved):

maxsam4:
Yeah. The function modifier thing was the last issue. 
Everything Looks good to me now.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +238,208 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
         }
         return Result.INVALID;
     }
 
+    /**
+     * @notice Provide the unlock amount for the given lockup for a particular user
+     */
+    function _getUnlockedAmountForLockup(LockUp[] userLockup, uint256 _lockupIndex) internal view returns (uint256) {
+            /*solium-disable-next-line security/no-block-members*/
+            if (userLockup[_lockupIndex].startTime > now) {
```
adamdossa: Maybe we should also be checking if:
`(userLockup[_lockupIndex].startTime.add(userLockup[_lockupIndex].lockUpPeriodSeconds) <= now)`
as an optimisation?

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +238,208 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
-        // check that totalPeriods evenly divides totalAmount
-        uint totalPeriods = lockUpPeriodSeconds.div(releaseFrequencySeconds);
+        // make sure the amount to be released per period is not too granular for the token
+        uint256 totalPeriods = _lockUpPeriodSeconds.div(_releaseFrequencySeconds);
+        uint256 amountPerPeriod = _lockupAmount.div(totalPeriods);
         require(
-            totalAmount % totalPeriods == 0,
-            "The total amount being locked up must be evenly divisible by the number of total periods"
+            amountPerPeriod.mul(totalPeriods) == _lockupAmount,
+            "lockup amount should be completely divisible by the amount per period"
```
adamdossa: Wondering how necessary this check is? Suppose that the amountPerPeriod is not a whole number (e.g. 10.5 tokens released per period) this would just effectively mean that the user has to keep a balance of 11 tokens to satisfy the clause.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
```diff
@@ -229,164 +238,208 @@ contract LockupVolumeRestrictionTM is ITransferManager {
...
         );
-
-        // make sure the amount to be released per period is not too granular for the token
-        uint amountPerPeriod = totalAmount.div(totalPeriods);
         require(
             amountPerPeriod % ISecurityToken(securityToken).granularity() == 0,
```
adamdossa: Wondering if it really matters if this is not true (amount per period is at a different granularity than the token)? Main concern is that it could be a bit confusing for issues  to have to ensure all these conditions are met (which could be hard if granularity is 1, and you're locking a prime number amount of tokens...)

adamdossa (changes requested):

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
maxsam4: Without this check, rounding off can cause tokens to be locked forever. Consider this example:
Granularity of token = 1
Total locked amount = 10
Number of periods = 6

This means, the amount per lock period = 10/6 = 1.67 = 1.
unlocked amount even after all periods are passed = 6*1 = 6
10 - 6 = 4 tokens are locked forever.

4 is a very small number as 1 token is 10**18 but it can cause amount mismatch and confusion as it will make transactions revert when user will try to transfer all of locked tokens.

That being said, I don't think we should be checking against granularity for amount released per period. We are never taking tokens away from the guy so the granularity issue is not a big problem for individual release amounts.

adamdossa: I guess this would need to be combined with the comment above that ignores any lockups which are past their end time. I feel like we should be doing this anyway for gas purposes.

contracts/modules/Experimental/TransferManager/LockupVolumeRestrictionTM.sol
satyamakgec: Suppose if granularity is 1 for the token and I passed prime no. as the lockup Amount with a 4 period then if I remove this check it will create the lockup for the 1 token and doesn't allow to user to withdrawal until full lockup gets passed. so prime nos are blocked for token whose granularity is one but allowed whose granularity is 18. i.e why I add this check to make sure that type of lockups will never be created

satyamakgec: gotcha what you are trying to say here.

adamdossa (approved):

satyamakgec:
@maxsam4 @adamdossa could you please review it again because of the data structure changes

contracts/modules/TransferManager/LockUpTransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+        } else {
+            // Calculate the no. of periods for a lockup 
+            uint256 noOfPeriods = (lockups[_lockupName].lockUpPeriodSeconds).div(lockups[_lockupName].releaseFrequencySeconds);
+            // Calculate the transaction time lies in which period
+            /*solium-disable-next-line security/no-block-members*/
+            uint256 elapsedPeriod = (now.sub(lockups[_lockupName].startTime)).div(lockups[_lockupName].releaseFrequencySeconds);
+            // Calculate the allowed unlocked amount per period
+            uint256 amountPerPeriod = (lockups[_lockupName].lockupAmount).div(noOfPeriods);
+            // Find out the unlocked amount for a given lockup
+            uint256 unLockedAmount = elapsedPeriod.mul(amountPerPeriod);
```
maxsam4: For better rounding off, we can first multiply and then divide. i.e
`unLockedAmount = lockups[_lockupName].lockupAmount.mul(elapsedPeriod).div(noOfPeriods);`

contracts/modules/TransferManager/LockUpTransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+        uint256 _lockupAmount,
+        uint256 _startTime,
+        uint256 _lockUpPeriodSeconds,
+        uint256 _releaseFrequencySeconds,
+        bytes32 _lockupName
+    )
+        internal
+    {
+        /*solium-disable-next-line security/no-block-members*/
+        require(lockups[_lockupName].startTime > now, "Not allowed to modify");
```
maxsam4: Why can't we allow modifying already started lockups?
I don't think it will cause any problems.

contracts/modules/TransferManager/LockUpTransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+        require(_startTime >= now, "Past time is not allowed");
```
maxsam4: I don't see a reason to not allow past lockups.
If we remove this check, we can make lockups "instant".
If we have this check then when the investor is creating a lock, they will have to enter start time a few minutes from now to allow mining time.

contracts/modules/TransferManager/LockUpTransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+    ) 
+        internal 
+        view 
+    {   
+        require(_lockUpPeriodSeconds != 0, "lockUpPeriodSeconds cannot be zero");
+        require(_releaseFrequencySeconds != 0, "releaseFrequencySeconds cannot be zero");
+        require(_lockupAmount != 0, "lockupAmount cannot be zero");
+
+        // check that the total amount to be released isn't too granular
+        require(
```
maxsam4: As we now have a condition that sets unlocked amount = locked amount after expiry, we don't need this restriction.

contracts/modules/TransferManager/LockUpTransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+            _lockupAmount % ISecurityToken(securityToken).granularity() == 0,
+            "The total amount to be released is more granular than allowed by the token"
+        );
+
+        // check that releaseFrequencySeconds evenly divides lockUpPeriodSeconds
+        require(
```
maxsam4: As we now have a condition that sets unlocked amount = locked amount after expiry, we don't need this restriction.

maxsam4 (changes requested):

contracts/modules/TransferManager/LockUpTransferManager.sol
satyamakgec: Yes, it is fine to give upper bound time to mine. otherwise, Issuer will able to create 6 months back lockup or more.

maxsam4: What is the problem with issuer creating a lockup that started 6 months before?

contracts/modules/TransferManager/LockUpTransferManager.sol
satyamakgec: Because a lockup type can be associated with multiple addresses. one change can affect all the address which are associated with a given lockup. that may not be viable during the running of multiple lockups. Which may not be the intention of the issuer to affect all lockups. 

To change a lockup for a particular user, Issuer can remove the previous lockup for him and create a one.
I can create an extra function that will allow modifying the lockup for all holders associated with it with where there is no start time check.

maxsam4: >one change can affect all the address which are associated with a given lockup. that may not be viable during the running of multiple lockups. Which may not be the intention of the issuer to affect all lockups.

It's pretty clear that one lockup can be applicable to multiple people. I don't think it will cause any confusions. If an issuer wants different lockups for different then he should not be using the same lockup.

contracts/modules/TransferManager/LockUpTransferManager.sol
satyamakgec: It doesn't make sense. If I am the user, I was able to transfer the tokens freely from last 6 months but on the Issuer portal, it will say the user was restricted from last 6 months while he already sells more than the lockup amount freely.

maxsam4: If an issuer is adding a lockup in past, he will be well aware of this. 

Anyway, how about we treat `_startTime = 0` as `_startTime = now` so that if an issuer wants to immediately start a lockup, he can do so. It will solve the UX problem I mentioned above.

satyamakgec: we could do that

maxsam4 (approved):

PR #410, opened 2018-11-13
Signed transfer manager fixed

comeonbuddy:
- [x ] The commit message follows our Submission guidelines
- [x ] Tests for the changes have been added (for bug fixes / features)

fixed signed transfer manager and test filess

current signed transfer manager was not working due to a pending issue we have double pre fix when signing messages because our version of web3 automatically adds one.

### What is the new behavior?
fixed the issue so the test file work

no


contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,169 @@
...
+    * @param _from address transfer from
+    * @param _to address transfer to
+    * @param _amount transfer amount
+    * @param _data signature
+    * @param _isTransfer bool value of isTransfer
+    * Sig needs to be valid (not used or deemed as invalid)
+    * Signer needs to be in the signers mapping
+    */
+    function verifyTransfer(address _from, address _to, uint256 _amount, bytes _data , bool _isTransfer) public returns(Result) {
+        if (!paused) {
```
adamdossa: You need the line:
`require(_isTransfer == false || msg.sender == securityToken, "Sender is not the owner");`
at the top of this function.

Otherwise anyone can directly call `verifyTransfer` and invalidate the signature without an actual transfer taking place.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,169 @@
...
+            
+            require(invalidSignatures[_data] != true, "Invalid signature - signature is either used or deemed as invalid");
+            bytes32 hash = keccak256(abi.encodePacked(this, _from, _to, _amount));
+            bytes32 _hash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
```
adamdossa: Can we leave _vars for parameters and call this something else (e.g. prependedHash)

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,169 @@
...
+            address signer = _recoverSignerAdd(_hash, _data);
+            require(signers[signer] == true, "Invalid signature - signer is not on the list");
```
adamdossa: I think that if the data doesn't match (i.e. `signers[signer] != true`) we should return Result.NA rather than throw. There could be another TM which allows the transfer anyway or interprets the data differently.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,169 @@
...
+    * @param _data signature
+    * Sig needs to be valid (not used or deemed as invalid)
+    * Signer needs to be in the signers mapping
+    */
+    function invalidSignature(address _from, address _to, uint256 _amount, bytes _data) public {
+        require(signers[msg.sender] == true, "Only signer is allowed to invalid signature.");
+        require(invalidSignatures[_data] != true, "This signature is invalid.");
+
+        bytes32 hash = keccak256(abi.encodePacked(this, _from, _to, _amount));
+        bytes32 _hash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
```
adamdossa: Same comment as above re. variable naming.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,169 @@
...
+
+        // return signer;
+        require(_recoverSignerAdd(_hash,_data) == msg.sender, "Incorrect Signer for this signature");
+
+        invalidSignatures[_data] = true;
+        emit InvalidSignature(_data);
+    }
+
+    // function invalidSignature(address _from, address _to, uint256 _amount, bytes _data) public view returns(address) {
```
adamdossa: Let's get rid of this commented out function if not needed

adamdossa (changes requested):

contracts/modules/TransferManager/SignedTransferManager.sol
comeonbuddy: I am a bit confused, current tests are assuming the istransfer = true to pass... then we invalid the sig.. do we require this to be false or true?

adamdossa: _isTransfer can be false or true, but if it is true then the only thing making the contract call should be the security token.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,158 @@
...
+        if (!paused) {
+
+            require(_isTransfer == false || msg.sender == securityToken, "Sender is not the owner");
+            
+            require(invalidSignatures[_data] != true, "Invalid signature - signature is either used or deemed as invalid");
+            bytes32 hash = keccak256(abi.encodePacked(this, _from, _to, _amount));
+            bytes32 prependedHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
+            address signer = _recoverSignerAdd(prependedHash, _data);
+            
+            require(signers[signer] == true, "Invalid signature - signer is not on the list");
```
adamdossa: Should get rid of:
`require(signers[signer] == true, "Invalid signature - signer is not on the list");`

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,158 @@
...
+            
+            if (signers[signer] != true){
+                return Result.NA;
+            } else {
+                invalidSignatures[_data] = true;
```
adamdossa: Only invalidate the signature if _isTransfer is true

adamdossa (changes requested):

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+    * @notice function to remove or add signer(s) onto the signer mapping
+    * @param _signers address array of signers
+    * @param _signersStats bool array of signers stats
+    */
+    function updateSigners(address[] _signers, bool[] _signersStats) public withPerm(ADMIN) {
+        require(_signers.length == _signersStats.length, "input array length does not match");
+        for(uint8 i=0; i<_signers.length; i++){
```
maxsam4:
```suggestion
        for(uint256 i=0; i<_signers.length; i++){
```
uin8 are more expensive than uint256 when not packed and also more vulnerable to accedental overflow.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,158 @@
...
+            
+            require(invalidSignatures[_data] != true, "Invalid signature - signature is either used or deemed as invalid");
+            bytes32 hash = keccak256(abi.encodePacked(this, _from, _to, _amount));
+            bytes32 prependedHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
+            address signer = _recoverSignerAdd(prependedHash, _data);
+            
+            require(signers[signer] == true, "Invalid signature - signer is not on the list");
+            
+            if (signers[signer] != true){
+                return Result.NA;
```
maxsam4: Shouldn't it return `Result.INVALID` ?

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,158 @@
...
+            } else {
+                invalidSignatures[_data] = true;
+                return Result.VALID;
```
maxsam4: Shouldn't it return `Result.NA` ?

test/helpers/signData.js
```diff
@@ -16,6 +17,38 @@ function signData(tmAddress, investorAddress, fromTime, toTime, expiryTime, rest
...
+        .slice(2);
+    packedData = new Buffer(packedData, "hex");
+    packedData = Buffer.concat([new Buffer(`\x19Ethereum Signed Message:\n${packedData.length.toString()}`), packedData]);
+    packedData = web3.sha3(`0x${packedData.toString("hex")}`, { encoding: "hex" });
+
+    return web3.eth.sign(account, packedData);
+}
+
+// test
+function signDataVerifyTransfer2 (tmAddress, fromAddress, toAddress, amount, pk) {
```
maxsam4: Remove this if you don't need it.

test/y_signed_transfer_manager.js
```diff
@@ -0,0 +1,323 @@
...
+        it("should fail to change signers stats without permission", async () => {
+            await catchRevert(I_SignedTransferManager.updateSigners([account_investor3], [false], {from: account_investor2}));
+        });
+
+        // it("should not be able to invalid siganture if not in the signer list", async () => {
+
+        //     console.log("pk is "+token_owner_pk);
+
+        //     const sig = signDataVerifyTransfer(
+        //         I_SignedTransferManager.address,
```
maxsam4: Remove unnecessary code.

test/y_signed_transfer_manager.js
```diff
@@ -0,0 +1,323 @@
...
+        it("should allow transfer with valid sig", async () => {
+            const sig = signDataVerifyTransfer(
+                I_SignedTransferManager.address,
+                account_investor1,
+                account_investor2,
+                web3.utils.toWei("1", "ether"),
+                account_investor3
+            );
+           
+            await I_SignedTransferManager.verifyTransfer(account_investor1, account_investor2, web3.utils.toWei("1", "ether"), sig, false, {from: account_investor3});
```
maxsam4: Don't directly call `I_SignedTransferManager.verifyTransfer`. Add Signed Transfer Manager to the Security token and try doing actual transfers.
Keep a couple direct calls with `_isTransfer = false` as well to test that functionality.

maxsam4 (changes requested):

contracts/modules/TransferManager/SignedTransferManager.sol
comeonbuddy: No - I think NA is OK. The signature may not be valid for this specific TM, but could be used elsewhere.

maxsam4: It's a POC module so it's fine I guess. I had a different vision in mind. 
In my vision, this module worked kind of like manual approval module where transfers will happen only when approved off chain by someone who is authorized.

contracts/modules/TransferManager/SignedTransferManager.sol
maxsam4: If we return VALID here, It will override the whitelist managers. Again, this is a POC module so it can be a whitelist module in itself.

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -78,6 +78,7 @@ contract SignedTransferManager is ITransferManager {
         if (!paused) {
 
             require(_isTransfer == false || msg.sender == securityToken, "Sender is not the owner");
+            require(_data.length != 0, "Invalid data input");
```
maxsam4: Don't revert. 
Return Result.INVALID

contracts/modules/TransferManager/SignedTransferManager.sol
```diff
@@ -78,6 +78,7 @@ contract SignedTransferManager is ITransferManager {
...
             
             require(invalidSignatures[_data] != true, "Invalid signature - signature is either used or deemed as invalid");
```
maxsam4: Don't revert.
Return Result.INVALID

contracts/modules/TransferManager/SignedTransferManager.sol
comeonbuddy: what do you think about this @adamdossa ? the original script returns VALID.

adamdossa: I think VALID is fine. Use-case I had in mind is an authorised entity (e.g. a signer) can provide data to whitelist a specific transaction. A bit like the MATM but using "off-chain" data instead of an on-chain list.

adamdossa (approved):

maxsam4 (approved):

maxsam4:
Closing in favor of https://github.com/PolymathNetwork/polymath-core/pull/533

PR #422, opened 2018-11-20
Vesting escrow wallet

kostind:

Added VestingEscrowWallet module.

### What is the new behavior?

No

### Any Other information:
Shouldn't be merged before all tests will be fixed


contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+
+    event AddSchedule(
+        address _beneficiary,
+        uint256 _numberOfTokens,
+        uint256 _duration,
+        uint256 _frequency,
+        uint256 _startTime,
+        uint256 _timestamp
+    );
+    event EditSchedule(
```
satyamakgec: use `modify` instead of the `Edit`

satyamakgec: and indexed the `_beneficiary`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    public
+    Module(_securityToken, _polyAddress)
+    {
+    }
+
+    /**
+     * @notice Function used to initialize the different variables
+     * @param _treasury Address of the treasury
+     */
+    function configure(address _treasury, address _token) public onlyFactory {
```
satyamakgec: we don't need an explicit token address. It always be a securityToken address that you can access directly as it is provided in the construction

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+
+    struct Template {
+        uint256 numberOfTokens;
+        uint256 duration;
+        uint256 frequency;
+    }
+
+    enum State {CREATED, STARTED, COMPLETED}
+
+    ERC20 public token;
```
satyamakgec: we don't need that you can directly use the `ISecurityToken` interface

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    function getInitFunction() public pure returns (bytes4) {
+        return bytes4(keccak256("configure(address,address)"));
+    }
+
+    /**
+     * @notice Used to deposit tokens from treasury
+     */
+    function depositTokens(uint256 _numberOfTokens) external withPerm(ADMIN) {
+        require(_numberOfTokens > 0, "Should be greater than zero");
+        token.safeTransferFrom(treasury, this, _numberOfTokens);
```
satyamakgec: We are not entertaining the simple ERC20 to hold into this wallet. we are only holding the ST init so use the `ISecurityToken` interface everywhere (or IERC20 for accessing the basic ERC20 functionality of the ST). we will use the ERC20 only on the reclaiming tokens as there will be a possibility where the issuer or anybody sends some simple ERC20 token into the contract. So we can extract those tokens from the contract

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+     * @param _numberOfTokens number of tokens
+     * @param _duration vesting duration
+     * @param _frequency vesting frequency
+     */
+    function addTemplate(uint256 _numberOfTokens, uint256 _duration, uint256 _frequency) external withPerm(ADMIN) {
+        _validateTemplate(_numberOfTokens, _duration, _frequency);
+        Template memory template;
+        template.numberOfTokens = _numberOfTokens;
+        template.duration = _duration;
+        template.frequency = _frequency;
```
satyamakgec: use a one-liner 
`templates.push(Tempalte(_numberOfTokens, _duration, _frequency));`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        templates.push(template);
+        /*solium-disable-next-line security/no-block-members*/
+        emit AddTemplate(_numberOfTokens, _duration, _frequency, now);
```
satyamakgec: emit the `index`  of the template if you are going to use the index further for any operation. no need to emit `now` as it automatically attaches with the events and we can read it from the blockchain until unless you are filtering according to timestamp

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    /**
+     * @notice Removes template
+     * @param _index index of the template
+     */
+    function removeTemplate(uint256 _index) external withPerm(ADMIN) {
+        require(_index < templates.length, "Template not found");
+        templates[_index] = templates[templates.length - 1];
+        templates.length--;
+        /*solium-disable-next-line security/no-block-members*/
+        emit RemoveTemplate(_index, now);
```
satyamakgec: same remove now

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+    /**
+     * @notice Returns count of templates
+     * @return count of templates
+     */
+    function getTemplateCount() external view withPerm(ADMIN) returns(uint256) {
```
satyamakgec: why we need `withPerm(ADMIN)` in it?. It is a simple getter.

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        emit DepositTokens(_numberOfTokens, now);
+    }
+
+    /**
+     * @notice Sends unassigned tokens to treasury
+     */
+    function sendToTreasury() external withPerm(ADMIN) {
+        uint256 amount = unassignedTokens;
+        unassignedTokens = 0;
+        token.safeTransfer(treasury, amount);
```
satyamakgec: Same here

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        emit AddTemplate(_numberOfTokens, _duration, _frequency, now);
+    }
+
+    /**
+     * @notice Removes template
+     * @param _index index of the template
+     */
+    function removeTemplate(uint256 _index) external withPerm(ADMIN) {
+        require(_index < templates.length, "Template not found");
+        templates[_index] = templates[templates.length - 1];
```
satyamakgec: you can save overwriting when `index == templates.length - 1`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        _validateSchedule(_beneficiary, _numberOfTokens, _duration, _frequency, _startTime);
+        require(_numberOfTokens <= unassignedTokens, "Not enough tokens");
+
+        Schedule memory schedule;
+        unassignedTokens = unassignedTokens.sub(_numberOfTokens);
+        schedule.numberOfTokens = _numberOfTokens;
+        schedule.lockedTokens = _numberOfTokens;
+        schedule.duration = _duration;
+        schedule.frequency = _frequency;
+        schedule.startTime = _startTime;
```
satyamakgec: you can use the one-liner here as well

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        schedule.nextTime = _startTime.add(schedule.frequency);
```
satyamakgec: what is the use of this you can calculate this into the fly

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+
+    /**
+     * @notice Adds vesting schedules from template for each of beneficiary
+     * @param _beneficiary beneficiary's addresses
+     * @param _index index of the template
+     * @param _startTime vesting start time
+     */
+    function addScheduleFromTemplate(address _beneficiary, uint256 _index, uint256 _startTime) public withPerm(ADMIN) {
+        require(_index < templates.length, "Template not found");
+        Template storage template = templates[_index];
```
satyamakgec: you can use memory here as you are not changing the value of the Template. and function could be external

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        token.safeTransfer(treasury, amount);
+        /*solium-disable-next-line security/no-block-members*/
+        emit SendToTreasury(amount, now);
+    }
+
+    /**
+     * @notice Sends available tokens to beneficiary
+     * @param _beneficiary beneficiary's address
+     */
+    function sendAvailableTokens(address _beneficiary) public withPerm(ADMIN) {
```
satyamakgec: rename the function to `pushAvailableTokens()`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        uint256 nextTime;
+        State state;
+    }
+
+    struct Data {
+        uint256 index;
+        Schedule[] schedules;
+        uint256 availableTokens;
+        uint256 claimedTokens;
+    }
```
satyamakgec: I don't know why you make the data structure complex. it could be very simple something like that


        struct Schedule {
                 uint256 totalAmount;
                 uint256 claimedTokens;
                 uint256 duration;
                 uint256 frequency;
                 uint256 startTime;
           }
          no need of state as there is no operation depend on the state.
          mapping(address => Schedule[]); 

nothing else

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        uint256 _duration,
+        uint256 _frequency,
+        uint256 _startTime
+    )
+        public
+        withPerm(ADMIN)
+    {
+        _validateSchedule(_beneficiary, _numberOfTokens, _duration, _frequency, _startTime);
+        require(_numberOfTokens <= unassignedTokens, "Not enough tokens");
+
```
satyamakgec: we can save one call, using this
      
        if (_numberOfTokens <= unassignedTokens) {
                  _depositTokens() -- internal version of the depositTokens
          }
This will avoid explicit calling of `depositTokens()`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    /**
+     * @notice Edits vesting schedules for each of beneficiary
+     * @param _beneficiary beneficiary's addresses
+     * @param _index index of schedule
+     * @param _numberOfTokens number of tokens
+     * @param _duration vesting duration
+     * @param _frequency vesting frequency
+     * @param _startTime vesting start time
+     */
+    function editSchedule(
```
satyamakgec: rename to `modifySchedule()`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        _validateSchedule(_beneficiary, _numberOfTokens, _duration, _frequency, _startTime);
+        require(_index < dataMap[_beneficiary].schedules.length, "Schedule not found");
+        Schedule storage schedule = dataMap[_beneficiary].schedules[_index];
+        /*solium-disable-next-line security/no-block-members*/
+        require(now < schedule.startTime, "Schedule started");
+        if (_numberOfTokens <= schedule.lockedTokens) {
+            unassignedTokens = unassignedTokens.add(schedule.lockedTokens - _numberOfTokens);
+        } else {
+            require((_numberOfTokens - schedule.lockedTokens) <= unassignedTokens, "Not enough tokens");
+            unassignedTokens = unassignedTokens.sub(_numberOfTokens - schedule.lockedTokens);
```
satyamakgec: Same here we can save one call by calling `_depositToken()` here

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+        }
+        schedule.numberOfTokens = _numberOfTokens;
+        schedule.lockedTokens = _numberOfTokens;
+        schedule.duration = _duration;
+        schedule.frequency = _frequency;
+        schedule.startTime = _startTime;
+        schedule.nextTime = _startTime.add(schedule.frequency);
```
satyamakgec: can do in a one-liner

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+     */
+    function revokeSchedule(address _beneficiary, uint256 _index) external withPerm(ADMIN) {
+        require(_beneficiary != address(0), "Invalid address");
+        require(_index < dataMap[_beneficiary].schedules.length, "Schedule not found");
+        Schedule[] storage schedules = dataMap[_beneficiary].schedules;
+        unassignedTokens = unassignedTokens.add(schedules[_index].lockedTokens);
+        schedules[_index] = schedules[schedules.length - 1];
+        schedules.length--;
+        if (schedules.length == 0) {
+            _revokeSchedules(_beneficiary);
```
satyamakgec: If issuer revoking the schedule then it also transfers all the unlocked tokens as well. Suppose you as an Issuer added me in a vesting schedule with a one year cliff and I left your company and did not pull my tokens yet but you revoke my schedule. then there is no way I can get my tokens. So I prefer whenever the issuer revoking the schedule function will check if there are any available tokens to send, it first transfers those tokens and then revokes the  schedule.

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+    function getAvailableTokens(address _beneficiary) external view returns(uint256) {
+        require(_beneficiary != address(0));
+        return dataMap[_beneficiary].availableTokens;
+    }
+
+    /**
+     * @notice Used to bulk send available tokens for each of beneficiaries
+     * @param _beneficiaries array of beneficiary's addresses
+     */
+    function batchSendAvailableTokens(address[] _beneficiaries) external withPerm(ADMIN) {
```
satyamakgec: rename the function to `pushAvailableTokenMutil()`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+
+    /**
+     * @notice Used to bulk add vesting schedules for each of beneficiaries
+     * @param _beneficiaries array of beneficiary's addresses
+     * @param _numberOfTokens number of tokens
+     * @param _duration vesting duration
+     * @param _frequency vesting frequency
+     * @param _startTime vesting start time
+     */
+    function batchAddSchedule(
```
satyamakgec: rename to AddScheduleMulti

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: add sanity check require(_beneficiaries.length > 1);

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,525 @@
...
+
+    /**
+     * @notice manually triggers update outside for beneficiary's schedule (can be used to reduce user gas costs)
+     * @param _beneficiary beneficiary's address of the schedule
+     */
+    function update(address _beneficiary) external withPerm(ADMIN) {
+        _update(_beneficiary);
+    }
+
+    function _update(address _beneficiary) private {
```
satyamakgec: What is the use of this function?

satyamakgec (changes requested):

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: Shouldn't we be able to work with any ERC20 tokens for Vesting Schedule?

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: We can for the first period.
eaxmple:
startTime  = 1st January 2018
duration = 3 years
frequency = 1 year

nextTime = 1st January 2019 (startTime + frequency)
nextTime = 1st January 2020 (nextTime + frequency)
nextTime = 1st January 2021 (nextTime + frequency)

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: I've added state because we it in use case for Employee Vesting Dapp:
Employees want to see the status of the vesting schedules (i.e. to start, vesting, completed, terminated)

kostind: I agree that state is redundant for now.

kostind: I can change struct Schedule and get rid of struct Data, but in this case we will need to calculate availableTokens for some beneficiary on the fly. Is it ok?

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: Shouldn't we prevent everyone to sending tokens to a treasury?
If someone will constantly invoke this method it will be hard to create a schedule.

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: No, now it is a module so it always bonded with the particular ST.

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: I am pointing here related to `safeTransfer`(use ST interface not ERC20). not related to permissions, permissions are good

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: but you can use `now` as well in calculation, something like that 
`period =  (now.sub(startTime)).div(frequency);` where a period should not be greater than duration.
and according to the period you can release the tokens.

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: yes to calculate the tokens you can create some view function. and you can also create a view function to know the state of the schedule. no need to store explicitly and update everytime

              function getScheduleState(address _user, uint256 _index) external view returns(State) {
                         Schedule memory _userSchedule = schedules[_user][0];
                         if (now < _userSchedule.startTime)
                                return State.NOTSTARTED;
                         else if (now > _userSchedule.startTime && now < _userScheduleTime.startTime.add(duration) )
                                 return State.RUNNING
                          .
                          .
                          .
             }

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: It's possible to pull available tokens by a beneficiary even if schedule is revoked.
It was one of the reason for complex structures: 
`
struct Data {
        uint256 index;
        Schedule[] schedules;
        uint256 availableTokens;
        uint256 claimedTokens;
    }
`    
Anyway I will simplify sctructure and add pushing tokens in revoking time.

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: Removed. Left only updateAll.

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: All changes have been done.
One difference: we need 3 variables with token numbers instead of 2.

    struct Schedule {
        uint256 numberOfTokens;
        uint256 releasedTokens;
        uint256 availableTokens;
        uint256 duration;
        uint256 frequency;
        uint256 startTime;
    }

numberOfTokens - number of tokens in creation time
releasedTokens - already released amount of tokens by schedule
availableTokens - available for withdrawing tokens

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+    mapping (address => Schedule[]) public schedules;
+    address[] public beneficiaries;
+
+    Template[] public templates;
+
+    event AddSchedule(
+        address indexed _beneficiary,
+        uint256 _numberOfTokens,
+        uint256 _duration,
+        uint256 _frequency,
```
satyamakgec: emit `_index` here as well

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+    /**
+     * @notice Used to deposit tokens from treasury
+     */
+    function depositTokens(uint256 _numberOfTokens) external withPerm(ADMIN) {
+        _depositTokens(_numberOfTokens);
+    }
+
+    function _depositTokens(uint256 _numberOfTokens) internal {
+        require(_numberOfTokens > 0, "Should be greater than zero");
+        ISecurityToken(securityToken).transferFrom(treasury, this, _numberOfTokens);
```
satyamakgec: no need to hard code the `from` to be the `treasury` it can be anyone (msg.sender). you can add `from` address in the event so it can be tracked

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+     * @param _startTime vesting start time
+     */
+    function addScheduleFromTemplate(address _beneficiary, uint256 _index, uint256 _startTime) external withPerm(ADMIN) {
+        _addScheduleFromTemplate(_beneficiary, _index, _startTime);
+    }
+
+    function _addScheduleFromTemplate(address _beneficiary, uint256 _index, uint256 _startTime) internal {
+        require(_index < templates.length, "Template not found");
+        Template memory template = templates[_index];
+        addSchedule(_beneficiary, template.numberOfTokens, template.duration, template.frequency, _startTime);
```
satyamakgec: you can also create the `_addSchedule()`  internal function to get rid of the `withPerm(ADMIN)` check for _addSchedule()

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+        public
+        withPerm(ADMIN)
+    {
+        _validateSchedule(_beneficiary, _numberOfTokens, _duration, _frequency, _startTime);
+        require(_index < schedules[_beneficiary].length, "Schedule not found");
+        Schedule storage schedule = schedules[_beneficiary][_index];
+        /*solium-disable-next-line security/no-block-members*/
+        require(now < schedule.startTime, "Schedule started");
+        if (_numberOfTokens <= schedule.numberOfTokens) {
+            unassignedTokens = unassignedTokens.add(schedule.numberOfTokens - _numberOfTokens);
```
satyamakgec: Try to use the SafeMath wherever you can
`unassignedTokens = unassignedTokens.add(schedule.numberOfTokens.sub(_numberOfTokens));`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+     * @notice Returns available tokens for beneficiary
+     * @param _beneficiary beneficiary's address
+     * @return available tokens for beneficiary
+     */
+    function getAvailableTokens(address _beneficiary) external view returns(uint256) {
+        return _getAvailableTokens(_beneficiary);
+    }
+
+    function _getAvailableTokens(address _beneficiary) internal view returns(uint256) {
+        require(_beneficiary != address(0));
```
satyamakgec: you can remove this require statement and move it `getAvailableTokens()`
and I think no need to create two functions make it one public function as its parameters also same

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+
+    /**
+     * @notice Revokes beneficiary's schedule
+     * @param _beneficiary beneficiary's address
+     * @param _index index of the schedule
+     */
+    function revokeSchedule(address _beneficiary, uint256 _index) external withPerm(ADMIN) {
+        require(_beneficiary != address(0), "Invalid address");
+        require(_index < schedules[_beneficiary].length, "Schedule not found");
+        _sendTokens(_beneficiary);
```
satyamakgec: In `_sendTokens(_beneficiary)` you are sending all the available tokens (i.e for every schedule whether it will be revoked or not). That seems wrong to me only try to find out the available tokens to given `_index`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+            _removeBeneficiary(_beneficiary);
+        }
+        emit RevokeSchedule(_beneficiary, _index);
+    }
+
+    /**
+     * @notice Revokes all beneficiary's schedules
+     * @param _beneficiary beneficiary's address
+     */
+    function revokeSchedules(address _beneficiary) public withPerm(ADMIN) {
```
satyamakgec: I prefer the name `revokeAllSchedules()` earlier names have a subtle difference that makes wrong function call

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+        return (
+            schedule.numberOfTokens,
+            schedule.duration,
+            schedule.frequency,
+            schedule.startTime,
+            _getScheduleState(_beneficiary, _index)
+        );
+    }
+
+    function _getScheduleState(address _beneficiary, uint256 _index) internal view returns(State) {
```
satyamakgec: no need to create the internal function here make it public so if I don't want all other details only want to query the state, I can

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+        uint256 _numberOfTokens,
+        uint256 _duration,
+        uint256 _frequency,
+        uint256 _startTime
+    )
+        external
+        withPerm(ADMIN)
+    {
+        for (uint256 i = 0; i < _beneficiaries.length; i++) {
+            addSchedule(_beneficiaries[i], _numberOfTokens, _duration, _frequency, _startTime);
```
satyamakgec: replace it by `_addSchedule()` internal function for addSchedule()

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+        uint256 _duration,
+        uint256 _frequency,
+        uint256 _startTime
+    )
+        external
+        withPerm(ADMIN)
+    {
+        require(_beneficiaries.length == _indexes.length, "Arrays sizes mismatch");
+        for (uint256 i = 0; i < _beneficiaries.length; i++) {
+            modifySchedule(_beneficiaries[i], _indexes[i], _numberOfTokens, _duration, _frequency, _startTime);
```
satyamakgec: can also create the `_modifySchedule()` internal version to get rid of the repetitive check of `withPerm(ADMIN)`

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+
+    /**
+     * @notice Used to bulk add vesting schedules for each of beneficiaries
+     * @param _beneficiaries array of beneficiary's addresses
+     * @param _numberOfTokens number of tokens
+     * @param _duration vesting duration
+     * @param _frequency vesting frequency
+     * @param _startTime vesting start time
+     */
+    function addScheduleMulti(
```
satyamakgec, line 440: There should be a new batch function as well where duration, noOfTokens, frequency can be different for different beneficiaries

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+                    if (schedule.releasedTokens < releasedTokens) {
+                        schedule.availableTokens = schedule.availableTokens.add(releasedTokens - schedule.releasedTokens);
+                        schedule.releasedTokens = releasedTokens;
+                    }
+                }
+            }
+        }
+    }
+
+    function _removeBeneficiary(address _beneficiary) internal {
```
satyamakgec: I don't think we need to call this everytime for the `revokeSchedule()` function calls. I prefer to make a trim function `trimBeneficiaries()` that will remove all the beneficiaries who have the `schedules.length = 0;` (save gas ). So whenever delegate/issuer wants to push the tokens it will call `trimBeneficiaries()` first and then it will call the `pushAvailableTokensToUser()`.

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
...
+            }
+            ISecurityToken(securityToken).transfer(_beneficiary, amount);
+            emit SendTokens(_beneficiary, amount);
+        }
+    }
+
+    /**
+     * @notice manually triggers update outside for all schedules (can be used to reduce user gas costs)
+     */
+    function updateAll() external withPerm(ADMIN) {
```
satyamakgec: As far my understanding `updateAll()` will be used to calculate the `availableTokens` for the beneficiaries. but It will make the execution cost more than it saved.  Suppose there are 5 users and every users period completion within the difference of 1 day (i.e U1 period completion on t1 then others will be t1 + 1, t1 + 2, t1 + 3 ...). then to make available tokens updated you will need to call the `updateAll()` function every day. that lead more cost burning instead of saving. I will recommend calculating the `available tokens` in the fly (at the execution time). you can check other modules implementation (BlackListTransferManager) we are doing this.

If you have any queries please let me know we can chat over slack or do a quick call.

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: > availableTokens - available for withdrawing tokens

can be calculated at execution time

satyamakgec (changes requested):
Please use `SafeMath` instead of direct addition and subtraction

contracts/modules/Wallet/VestingEscrowWallet.sol
```diff
@@ -0,0 +1,507 @@
+pragma solidity ^0.4.24;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
```
satyamakgec: no need of SafeERC20 remove it

kostind:
> availableTokens - can be calculated at execution time

Done

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: What about startTime?
Can we change this function to use arrays instead of single values?
Adding the same schedule for different beneficiaries is possible using templates.
We need to optimize contact size, adding new function produces "out of gas" error during contract creation.

contracts/modules/Wallet/VestingEscrowWallet.sol
kostind: Should I use msg.sender in sendToTreasury as well or pass treasury as a paramter?

satyamakgec: always use `msg.sender`

contracts/modules/Wallet/VestingEscrowWallet.sol
satyamakgec: Yes it is good, change it to arrays to provide different vesting schedules to different beneficiaries

adamdossa:
@kostind - Couple of small questions - overall module looks great.
1. Please can we wrap `transferFrom` calls in a `require` statement. Not strictly needed as the ST will throw if there is a failure, but good practice in case the ST implementation were to change in the future (and more in keeping with the ERC20 spec.)
2. Any reason not to allow the `sendToTreasury` function to take an explicit amount in case you only want to transfer some unassigned tokens back to treasury (the function can check that the amount requested is less than `unassignedTokens`?
3. To avoid the loop in e.g. `removeTemplate` we could store the index of the template into `templateNames` in the Template struct. I always worry about unbounded loops ;-).

kostind:
@adamdossa @satyamakgec All comments have been applied.

satyamakgec:
@kostind everything is good once the test cases will successfully be passed. I will approve the PR

satyamakgec (approved):

PR #431, opened 2018-11-22
[WIP] [CLI] PolymathRegistry class

shuffledex:


CLI/commands/ContractClass.js
```diff
@@ -0,0 +1,32 @@
+const common = require('./common/common_functions')
+const abis = require('./helpers/contract_abis')
+const gbl = require('./common/global')
+
+class Contract {
+  static async connect (contractName, address) {
```
VictorVicente: connect is not async anymore

CLI/commands/ContractClass.js
```diff
@@ -0,0 +1,32 @@
...
+    return new web3.eth.Contract(abis[contractName](), address)
+  }
+
+  static async call (contract, methodName, methodArguments) {
+    try {
+      return await contract.methods[methodName].apply(this, methodArguments).call()
+    } catch (e) {
+      console.log(`${methodName} call - `.toUpperCase(), e.message)
+      process.exit(0)
```
VictorVicente: We should avoid to exit here. We should discuss how we want to manage errors.

CLI/commands/ContractClass.js
```diff
@@ -0,0 +1,32 @@
...
+    }
+  }
+
+  static async owner (contract) {
+    return await contract.methods.owner().call() == Issuer.address
```
VictorVicente: owner should return the contract owner. iAmOwner() should make this comparison instead.

VictorVicente (changes requested):

VictorVicente:
We are facing the SDK from another perspective now.

PR #433, opened 2018-11-25
Optimise GTM

adamdossa:
- Add an Offset that can be used to move all from & to dates forwards or backwards by a fixed offset.
  - Add `address[] public investors` to record a list of all addresses that have been added to the whitelist
  - Fix for when `allowAllWhitelistIssuances` is FALSE
  - Clearer logic around toTime, fromTime & expiryTime
  - Make GTM a Proxy based implementation to reduce deployment gas costs


contracts/modules/ModuleStorage.sol
```diff
@@ -0,0 +1,30 @@
...
+ */
+contract ModuleStorage {
+
+    /**
+     * @notice Constructor
+     * @param _securityToken Address of the security token
+     * @param _polyAddress Address of the polytoken
+     */
+    constructor (address _securityToken, address _polyAddress) public {
+        securityToken = _securityToken;
```
maxsam4, line 17: missing data validation

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -168,16 +145,19 @@ contract GeneralTransferManager is ITransferManager {
                 //Anyone on the whitelist can transfer, regardless of time
                 return (_onWhitelist(_to) && _onWhitelist(_from)) ? Result.VALID : Result.NA;
```
maxsam4, line 151: Should we allow people not on the whitelist to transfer their tokens to someone on the whitelist?
This case will help st get back to the allowed state.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -168,16 +145,19 @@ contract GeneralTransferManager is ITransferManager {
...
             }
+            if (_from == issuanceAddress) {
+                return (_onWhitelist(_to) && _adjustTimes(whitelist[_to].toTime) <= uint64(now)) ? Result.VALID : Result.NA;
+            }
             //Anyone on the whitelist can transfer provided the blocknumber is large enough
             /*solium-disable-next-line security/no-block-members*/
-            return ((_onWhitelist(_from) && whitelist[_from].fromTime <= now) &&
-                (_onWhitelist(_to) && whitelist[_to].toTime <= now)) ? Result.VALID : Result.NA; /*solium-disable-line security/no-block-members*/
+            return ((_onWhitelist(_from) && _adjustTimes(whitelist[_from].fromTime) <= uint64(now)) &&
+                (_onWhitelist(_to) && _adjustTimes(whitelist[_to].toTime) <= uint64(now))) ? Result.VALID : Result.NA; /*solium-disable-line security/no-block-members*/
```
maxsam4, line 165: Shouldn't we be checking `whitelist[_to].toTime` AND `whitelist[_from].fromTime` for both `_from` and `_to`?
The current implementation will allow the transfer to happen when whitelist period of `_to` hasn't even started and whitelist period of `_from` has already expired.
Also, maybe we should remove forcing the sender to be on the whitelist.

contracts/modules/TransferManager/GeneralTransferManagerFactory.sol
```diff
@@ -21,6 +23,7 @@ contract GeneralTransferManagerFactory is ModuleFactory {
         description = "Manage transfers using a time based whitelist";
         compatibleSTVersionRange["lowerBound"] = VersionUtils.pack(uint8(0), uint8(0), uint8(0));
         compatibleSTVersionRange["upperBound"] = VersionUtils.pack(uint8(0), uint8(0), uint8(0));
+        logicContract = _logicContract;
```
maxsam4, line 27: Data validation checks missing.

maxsam4 (changes requested):

contracts/modules/TransferManager/GeneralTransferManager.sol
satyamakgec: > Shouldn't we be checking whitelist[_to].toTime AND whitelist[_from].fromTime for both _from and _to?

@maxsam4 we are checking it but it can be overwrite if any offset time is set by the issuer. 
 

> The current implementation will allow the transfer to happen when whitelist period of _to hasn't even started and whitelist period of _from has already expired.

As per the current implementation it is fine because whiteliste period is started when the user get whitelisted and remains whitelisted until its `expiryTime` is < `now`, that we are checking in the `_onWhiteList()` function. So both users will be in the white list and the second part is to check the buy lockup period of  `_to` and sell lockup period of `_from`. that we are covering in the `_adjustTimes(whitelist[_to].toTime) <= uint64(now))) ` &  _adjustTimes(whitelist[_from].fromTime) <= uint64(now)). 

> Also, maybe we should remove forcing the sender to be on the whitelist.

That can be done by assuming that  `_to` address is always be whitelisted.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -192,16 +172,38 @@ contract GeneralTransferManager is ITransferManager {
...
+    function _modifyWhitelist(
+        address _investor,
+        uint64 _fromTime,
+        uint64 _toTime,
+        uint64 _expiryTime,
+        uint8 _canBuyFromSTO
+    )
+        internal
+    {
+        if (whitelist[_investor].added == uint8(0)) {
```
satyamakgec, line 209: missing `require(_investor != address(0))`

satyamakgec (changes requested):

contracts/modules/TransferManager/GeneralTransferManager.sol
maxsam4: Ah, thanks for the explanation. Correct me if I'm wrong:
1) kyc expiry time is different from transfer restriction time. 
2) `toTime` and `fromTime` are not starting and ending time of allowance/restriction but times up till which an investor and `receive` or `send` tokens. 

We should consider adding a comment about it in the code

satyamakgec: Yes, you are right for both points,

1. KYC expiry time is more of validating time for your identity after expiration may be an investor turn from accredited to the non-accredited investor or something else because every specific provide by him/her are changeable by time.
2. `toTime`  and `fromTime` are `buy` and `sell` lockup. 

Yes description is need to be added in the code specifically at the declaration of `TimeRestriction`. while you can read this into the `modifywhitelist()` function params

contracts/modules/ModuleStorage.sol
adamdossa: The reason I removed this is that we need to deploy the logic contract (which needs to call it's inherited constructor in ModuleStorage) as a library where it doesn't make much sense to set these variables. Currently in `2_deploy_contracts.js` I just set these to 0x0. I could set to 0x1 instead and then enforce the non-0x0 constraint, but this is kind of enforced anyway as these contracts are generally deployed by the factory which has guarantees over these variables anyway.

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: I think probably not. If someone not on the whitelist has tokens (for whatever reason, e.g. a MATM approval) they shouldn't be able to transfer them out unless they've been defined in the GTM.

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: I added the following comment next to the struct definition for a bit more clarity:
```
    // _fromTime is the time from which the _investor can send tokens
    // _toTime is the time from which the _investor can receive tokens
    // if allowAllWhitelistIssuances is TRUE, then _toTime is ignored when receiving tokens from the issuance address
    // if allowAllWhitelistTransfers is TRUE, then _toTime and _fromTime is ignored when sending or receiving tokens
    // in any case, any investor sending or receiving tokens, must have a _expiryTime in the future
```

contracts/modules/TransferManager/GeneralTransferManagerFactory.sol
adamdossa: Will add

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Will add, although this is related to the question as to whether or not we should be whitelisting the issuance address (in this implementation not necessary, in the previous implementation it was).

maxsam4 (approved):

satyamakgec (approved):

PR #437, opened 2018-11-27
Multiple stable coins support in USD Tiered STO

maxsam4:

Added Multiple stable coins support in USD Tiered STO

1. An array of stable coin addresses has to be passed now rather than a single stable coin when configuring.
2. The investor has to provide the stable coin address they want to use when buying tokens with USD.


contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -149,8 +154,13 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
     }
 
-    modifier validDAI {
-        require(fundRaiseTypes[uint8(FundRaiseType.DAI)], "DAI not allowed");
+    modifier validSC {
+        require(fundRaiseTypes[uint8(FundRaiseType.SC)], "Stable coins not allowed");
+        _;
+    }
+
+    modifier validUSDToken(address _usdToken) {
```
satyamakgec: both modifier could be merge into one

adamdossa:
@maxsam4 Only minor comment is you could consider combining `validSC` and `validUSDToken` as I think they'll always be called together.

Otherwise looks good to me, with the caveat that the dApp / CLI will need some minor modifications for this.

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -319,16 +329,23 @@ contract USDTieredSTO is ISTO, ReentrancyGuard {
...
+        _modifyUSDTokens(_usdTokens);
+    }
+
+    function _modifyUSDTokens(address[] _usdTokens) internal {
+        for(uint256 i = 0; i < usdTokens.length; i++) {
+            usdTokenEnabled[usdTokens[i]] = false;
+        }
+        usdTokens = _usdTokens;
+        for(i = 0; i < _usdTokens.length; i++) {
+            usdTokenEnabled[_usdTokens[i]] = true;
```
satyamakgec, line 268: can add the 0x0 check of an address of usdTokens

satyamakgec (changes requested):

contracts/modules/STO/USDTieredSTO.sol
satyamakgec: that is already fixed

maxsam4:
Thanks @adamdossa 
Pushed commit to merge the modifiers. My coding style is still one function one task but that does not work with solidity. Adapting is hard :).

contracts/modules/STO/USDTieredSTO.sol
maxsam4: Added.
Thanks @satyamakgec

satyamakgec (approved):

CLI/commands/sto_manager.js
```diff
@@ -283,53 +284,110 @@ function fundingConfigUSDTieredSTO() {
...
+      },
+      limitMessage: "Must be a valid address",
+      defaultInput: Issuer.address
+    });
+    if (addresses.reserveWallet == "") addresses.reserveWallet = Issuer.address;
+  
+    let listOfAddress;
+  
+    if (usdTokenRaise) {
+      addresses.usdToken = readlineSync.question('Enter the address (or many addresses that you want separated by comma) of the USD Token or stable coin (' + usdToken.options.address + '): ', {
```
F-OBrien: I was just read this and it didn't sound correct so this is my suggested edit. Also "Stable coin" is used throughout and not "USD Token" so suggest removing it. No harm saying "USD stable coin" explicitly in case someone tries to use a different currency stable coin and as the Oracles are USD only. 
```suggestion
      addresses.usdToken = readlineSync.question('Enter the address (or multiple addresses separated by commas) of the USD stable coin(s) (' + usdToken.options.address + '): ', {
```

shuffledex: Thanks @F-OBrien, you have right!!! I will modify the string in the next commit.

maxsam4: @shuffledex I have merged the suggestion, please pull on your local.

PR #440, opened 2018-11-28
CappedSTO price fix

maxsam4:

Capped STO now allows the issuer to set the price of their token higher than 1ETH/token

The rate is now assumed to be multiplied by 10^18 by the smart contracts. So if a user wants to sell 1 unit of their token for 1 unit of poly, they need to set rate as 10^18 rather than 1.


contracts/modules/STO/CappedSTO.sol
```diff
@@ -49,7 +50,7 @@ contract CappedSTO is ISTO, ReentrancyGuard {
      * @param _startTime Unix timestamp at which offering get started
      * @param _endTime Unix timestamp at which offering get ended
      * @param _cap Maximum No. of tokens for sale
-     * @param _rate Token units a buyer gets per wei / base unit of POLY
+     * @param _rate Token units multiplied by 10^18 a buyer gets per wei / base unit of POLY
```
monitz87: If you're gonna specify this for this param, you should specify it for the cap as well (same for the function signature comment)

maxsam4: cap is not multiplied by 10^18. Consider indivisible tokens where decimals = 1. In that case, the cap will not be multiplied by 10^18. To sell max 2 tokens, I'll have to set the cap as 2 only.

cap is to be entered multiplied by 10^decimals like all values of tokens are entered in smart contracts usually.

In the comment regarding rate, the base unit of POLY means 10 to the power minus 18 POLY token. If a user wants 1 POLY = 1 Security Token even then they have to enter rate as 10^18.

To summarize: cap is multiplied according to the security token's decimals like every other number like transfer amount is. On the other hand, rate is multiplied by fixed 10^18.

That being said, you are right, I should make this clearer in the comments, thanks.

adamdossa (approved):
Looks good to me!

PR #443, opened 2018-11-28
Optimise GTM

adamdossa:
- Proxy deployments for GTM and Dividend modules (deployment gas optimisation)
  - Data packing for GTM (whitelist gas optimisation)
  - `getInvestors`, `getAllInvestorsData`, `getInvestorsData` added to GTM
  - `modifyDefaults(uint64 _defaultFromTime, uint64 _defaultToTime)` added which sets a default timestamp used when `fromTime` or `toTime` are 0
  - Fix for when `allowAllWhitelistIssuances` is FALSE (no requirement to whitelist issuance address)
  - `toTime` & `fromTime` both being 0 is interpreted consistently
  - version bumped to 2.0.0


contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
```diff
@@ -1,17 +1,16 @@
...
 import "./DividendCheckpoint.sol";
+import "./ERC20DividendCheckpointStorage.sol";
 import "../../interfaces/IOwnable.sol";
 import "../../interfaces/IERC20.sol";
 
 /**
  * @title Checkpoint module for issuing ERC20 dividends
  */
-contract ERC20DividendCheckpoint is DividendCheckpoint {
+contract ERC20DividendCheckpoint is ERC20DividendCheckpointStorage, DividendCheckpoint {
```
maxsam4, line 11: **Just a note** 
This changes the storage order from before. 
It does not matter as we aren't upgrading old contracts but if/when we do support upgrading, we'll have to be careful.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -144,7 +126,7 @@ contract GeneralTransferManager is ITransferManager {
         allowAllBurnTransfers = _allowAllBurnTransfers;
         emit AllowAllBurnTransfers(_allowAllBurnTransfers);
     }
-
+    event Times(uint64 _from, uint64 _to);
```
maxsam4: Seems like this slipped through the clean up.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -168,16 +150,21 @@ contract GeneralTransferManager is ITransferManager {
                 //Anyone on the whitelist can transfer, regardless of time
                 return (_onWhitelist(_to) && _onWhitelist(_from)) ? Result.VALID : Result.NA;
             }
+            if (_from == issuanceAddress && (whitelist[_to].canBuyFromSTO == 0) && _isSTOAttached()) {
```
maxsam4: Shouldnt it be`whitelist[_to].canBuyFromSTO != 0` rather than `whitelist[_to].canBuyFromSTO == 0`?

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -267,8 +289,14 @@ contract GeneralTransferManager is ITransferManager {
...
-        //Passing a _time == 0 into this function, is equivalent to removing the _investor from the whitelist
-        whitelist[_investor] = TimeRestriction(_fromTime, _toTime, _expiryTime, _canBuyFromSTO);
+        if (whitelist[_investor].added == uint8(0)) {
+            investors.push(_investor);
+        }
+        uint8 canBuyFromSTO = 0;
+        if (_canBuyFromSTO) {
+            canBuyFromSTO = 1;
+        }
+        whitelist[_investor] = TimeRestriction(uint64(_fromTime), uint64(_toTime), uint64(_expiryTime), canBuyFromSTO, uint8(1));
```
maxsam4: can we reuse `_modifyWhitelist` here?

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -200,12 +187,39 @@ contract GeneralTransferManager is ITransferManager {
...
+        uint64 _expiryTime,
+        uint8 _canBuyFromSTO
+    )
+        internal
+    {
+        require(_investor != address(0), "Invalid investor");
+        if (whitelist[_investor].added == uint8(0)) {
+            investors.push(_investor);
+        }
+        whitelist[_investor] = TimeRestriction(_fromTime, _toTime, _expiryTime, _canBuyFromSTO, uint8(1));
```
maxsam4: can we emit the event here rather than separately on the outside functions?

maxsam4 (dismissed):

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -76,6 +52,12 @@ contract GeneralTransferManager is ITransferManager {
     {
     }
```
pabloruiz55: Missing natspec comments

contracts/modules/Checkpoint/ERC20DividendCheckpointFactory.sol
```diff
@@ -24,6 +26,7 @@ contract ERC20DividendCheckpointFactory is ModuleFactory {
         description = "Create ERC20 dividends for token holders at a specific checkpoint";
         compatibleSTVersionRange["lowerBound"] = VersionUtils.pack(uint8(0), uint8(0), uint8(0));
         compatibleSTVersionRange["upperBound"] = VersionUtils.pack(uint8(0), uint8(0), uint8(0));
+        logicContract = _logicContract;
```
satyamakgec, line 31: shouldn't we add 0x0 check for logicContract ?

contracts/modules/TransferManager/GeneralTransferManager.sol
satyamakgec: 0 - false
1 - true 
so I think it is correct. By assuming the minting during the STO is performed by the MATM

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -200,12 +187,39 @@ contract GeneralTransferManager is ITransferManager {
...
     {
-        //Passing a _time == 0 into this function, is equivalent to removing the _investor from the whitelist
-        whitelist[_investor] = TimeRestriction(_fromTime, _toTime, _expiryTime, _canBuyFromSTO);
+        uint8 canBuyFromSTO = 0;
+        if (_canBuyFromSTO) {
+            canBuyFromSTO = 1;
+        }
+        _modifyWhitelist(_investor, uint64(_fromTime), uint64(_toTime), uint64(_expiryTime), canBuyFromSTO);
         /*solium-disable-next-line security/no-block-members*/
         emit ModifyWhitelist(_investor, now, msg.sender, _fromTime, _toTime, _expiryTime, _canBuyFromSTO);
```
satyamakgec: everything can be done(canBuyFromSTO check , event emission)  into  `_modifyWhitelist()` so you don't need to write the same thing in the `modifyWhitelistMulti()` again

satyamakgec (changes requested):

contracts/modules/Checkpoint/ERC20DividendCheckpointFactory.sol
maxsam4: It is checked in the proxy contract.

contracts/modules/TransferManager/GeneralTransferManager.sol
maxsam4: What's the purpose of `canBuyFromSTO` then?

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -168,16 +150,21 @@ contract GeneralTransferManager is ITransferManager {
                 //Anyone on the whitelist can transfer, regardless of time
                 return (_onWhitelist(_to) && _onWhitelist(_from)) ? Result.VALID : Result.NA;
             }
+            if (_from == issuanceAddress && (whitelist[_to].canBuyFromSTO == 0) && _isSTOAttached()) {
+                return Result.NA;
+            }
             if (allowAllWhitelistIssuances && _from == issuanceAddress) {
```
maxsam4: As a minor optimization, we can use nested if to test the 3 conditions that require `_from == issuanceAddress`

contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
adamdossa: Ack.

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Removed.

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Yep - done. Leads to weird gas estimates in ganache but this is fine so long as we remember it.

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Added

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Done

contracts/modules/TransferManager/GeneralTransferManager.sol
adamdossa: Done

contracts/modules/Checkpoint/ERC20DividendCheckpointFactory.sol
satyamakgec: I was preferring the check here because if the logic contract is `0x0` here then factory deployment is just a wastage of the gas because there is no way to change the `logicContract` in the  factory contract

satyamakgec (approved):

VictorVicente:
@adamdossa If we are going to redeploy both dividends modules, we should add them to the changelog and change their version too

PR #450, opened 2018-12-03
MATM changes

comeonbuddy:

- Removed all the blocking related functions and events in MATM
- Given a single address, anyone is able to see a list of all active approvals from/to that address
- Improved MATM to allow allowances and expiryTimes to be changed for existing approvals.
- As an issuer, you are able to add/revoke in batches (using a multi approach)
- As an issuer, you are able to add a description to a manual approval.
- Anyone is able to see a list of all active approvals

Yes


satyamakgec (approved):
Changes are looking fine to me.  
@comeonbuddy you don't need to remove  `/*solium-disable-next-line security/no-block-members*/` this from the code it only used for linting :)

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -86,14 +99,47 @@ contract ManualApprovalTransferManager is ITransferManager {
...
+    * @param _from is the address array from which transfers are approved
+    * @param _to is the address array to which transfers are approved
+    * @param _allowance is the array of approved amounts 
+    * @param _expiryTime is the array of the times until which eath transfer is allowed
+    * @param _description is the description array for these manual approvals
+    */
+    function addManualApprovalMulti(address[] _from, address[] _to, uint256[] _allowance, uint256[] _expiryTime, bytes32[] _description) public withPerm(TRANSFER_APPROVAL) {
+        require(_from.length == _to.length && _to.length == _allowance.length && _allowance.length == _expiryTime.length && _expiryTime.length == _description.length, "input array numbers not matching");
+        for(uint8 i=0; i<_from.length; i++){
+            addManualApproval(_from[i], _to[i], _allowance[i], _expiryTime[i], _description[i]);
```
maxsam4: Create an internal `_addManualApproval()` function and reuse it here and in `addManualApproval` function to save gas.

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -102,9 +148,44 @@ contract ManualApprovalTransferManager is ITransferManager {
     function revokeManualApproval(address _from, address _to) public withPerm(TRANSFER_APPROVAL) {
         require(_to != address(0), "Invalid to address");
         delete manualApprovals[_from][_to];
+
+        //find the record in active approvals array & delete it
+        uint256 index;
+        for(uint256 i = 0; i < activeManualApprovals.length; i++){
```
maxsam4: This will go out of gas when a lot of manual approvals are added. Consider saving the index in the struct or somewhere else to directly delete elements. See other places we are deleting restrictions for the pattern of moving last element to current position and deleting last element to preserve indexes.

maxsam4 (changes requested):
Still need to review logic. Only completed style review currently.

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -19,20 +19,19 @@ contract ManualApprovalTransferManager is ITransferManager {
...
-    }
-
-    //Manual blocking allows you to specify a list of blocked address pairs with an associated expiry time for the block
-    struct ManualBlocking {
-        uint256 expiryTime;
+        bytes32 description;
     }
 
     //Store mappings of address => address with ManualApprovals
     mapping (address => mapping (address => ManualApproval)) public manualApprovals;
```
maxsam4: Instead of maintaining this redundant struct mapping, can we just use `approvalsIndex` mapping to fetch index and then fetch required data from `activeManualApprovals` using that index?

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -19,20 +19,19 @@ contract ManualApprovalTransferManager is ITransferManager {
...
+        address from;
+        address to;
         uint256 allowance;
         uint256 expiryTime;
-    }
-
-    //Manual blocking allows you to specify a list of blocked address pairs with an associated expiry time for the block
-    struct ManualBlocking {
-        uint256 expiryTime;
+        bytes32 description;
```
maxsam4, line 26: is `bytes32` enough for the description? we can't even fit an IPFS hash in `bytes32`.

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -111,51 +101,233 @@ contract ManualApprovalTransferManager is ITransferManager {
...
-        manualBlockings[_from][_to] = ManualBlocking(_expiryTime);
-        emit AddManualBlocking(_from, _to, _expiryTime, msg.sender);
+        require(manualApprovals[_from][_to].allowance != 0, "Approval does not exists");
+        uint256 currentAllowance = manualApprovals[_from][_to].allowance;
+        uint256 newAllowance;
+        if (_change == 1) {
+            // Allowance get increased
+            manualApprovals[_from][_to].allowance = (
+            manualApprovals[_from][_to].allowance.add(_changedAllowance));
+            newAllowance = currentAllowance.add(_changedAllowance);
```
maxsam4, line 203: A minor optimization can be that we calculate `newAllowance` first and then set `manualApprovals[_from][_to].allowance = newAllowance`. The same optimization can also be done in the else if case below.

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -111,51 +101,233 @@ contract ManualApprovalTransferManager is ITransferManager {
...
     * @param _from is the address from which transfers are approved
     * @param _to is the address to which transfers are approved
     */
-    function revokeManualApproval(address _from, address _to) public withPerm(TRANSFER_APPROVAL) {
+    function revokeManualApproval(address _from, address _to) external withPerm(TRANSFER_APPROVAL) {
+        _revokeManualApproval(_from, _to);
+    }
+
+    function _revokeManualApproval(address _from, address _to) internal {
         require(_to != address(0), "Invalid to address");
```
maxsam4: This check seems redundant.

maxsam4 (changes requested):
The code looks good, Just some minor things that can be optimized.
Test cases need to be fixed/added.

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
satyamakgec: I already changed it to string during improving the test cases

satyamakgec: I think we have to keep the bytes32 as a small description length. Otherwise we can't call batch functions with variable length datatype.
string[], bytes[] are not allowed in 0.4.24;

contracts/modules/TransferManager/ManualApprovalTransferManager.sol
```diff
@@ -19,34 +19,34 @@ contract ManualApprovalTransferManager is ITransferManager {
...
-        uint256 expiryTime;
-    }
-
-    //Store mappings of address => address with ManualApprovals
-    mapping (address => mapping (address => ManualApproval)) public manualApprovals;
-
-    //Store mappings of address => address with ManualBlockings
-    mapping (address => mapping (address => ManualBlocking)) public manualBlockings;
+    mapping (address => mapping (address => uint256)) public approvalIndex;
+    mapping (address => mapping (address => bool)) public hasApproval;
```
maxsam4: Rather than maintaining `hasApproval`, we can just set approvalIndex as index + 1 and then if approvalIndex = 0, that means approval is not present. 

Setting hasApproval will take ~40k gas and it can add up to a lot when using multi functions.

maxsam4 (changes requested):

VictorVicente:
Edited title and description to include all new MATM changes

maxsam4 (approved):

maxsam4 (approved):

PR #459, opened 2018-12-06
change the type of proxy deployment for USDTieredSTO

satyamakgec:


contracts/modules/STO/ISTO.sol
```diff
@@ -3,30 +3,16 @@ pragma solidity ^0.4.24;
...
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
 /**
  * @title Interface to be implemented by all STO modules
  */
-contract ISTO is Module, Pausable  {
+contract ISTO is ISTOStorage, Module, Pausable  {
     using SafeMath for uint256;
 
     enum FundRaiseType { ETH, POLY, DAI }
```
maxsam4, line 15: Do enums affect storage?

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -7,86 +7,16 @@ import "../../RegistryUpdater.sol";
...
-    // Minimum investable amount in USD
-    uint256 public minimumInvestmentUSD;
-
-    // Final amount of tokens returned to issuer
-    uint256 public finalAmountReturned;
-
-    // Array of Tiers
-    Tier[] public tiers;
+    string public constant POLY_ORACLE = "PolyUsdOracle";
+    string public constant ETH_ORACLE = "EthUsdOracle";
```
maxsam4, line 19: Do `public constant` affect storage?
`internal constant` don't affect storage but I am not sure about `public constant`. 
They probably don't but we can never be careful enough.

contracts/modules/STO/USDTieredSTOFactory.sol
```diff
@@ -9,17 +10,17 @@ import "../../libraries/Util.sol";
...
      */
-    constructor (address _polyAddress, uint256 _setupCost, uint256 _usageCost, uint256 _subscriptionCost, address _proxyFactoryAddress) public
+    constructor (address _polyAddress, uint256 _setupCost, uint256 _usageCost, uint256 _subscriptionCost, address _logicContract) public
     ModuleFactory(_polyAddress, _setupCost, _usageCost, _subscriptionCost)
     {
-        require(_proxyFactoryAddress != address(0), "0x address is not allowed");
-        USDTieredSTOProxyAddress = _proxyFactoryAddress;
+        require(_logicContract != address(0), "0x address is not allowed");
+        logicContract = _logicContract;
         version = "1.0.0";
```
maxsam4: Let's bump the version?

contracts/proxy/USDTieredSTOProxy.sol
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+    * @notice Constructor
+    * @param _securityToken Address of the security token
+    * @param _polyAddress Address of the polytoken
+    * @param _implementation representing the address of the new implementation to be set
+    */
+    constructor (address _securityToken, address _polyAddress, address _implementation)
+    public
+    ModuleStorage(_securityToken, _polyAddress)
```
maxsam4, line 23: Need to initialize `OwnedProxy` as well using `OwnedProxy()`

maxsam4: @adamdossa I just noticed that we haven't initialized it in our other proxy modules either. We need to fix this.
We may also consider refactoring owned proxy to use SecurityToken.owner().

maxsam4 (changes requested):

contracts/modules/STO/ISTO.sol
satyamakgec: As per my knowledge nope, they are only the declaration

contracts/modules/STO/USDTieredSTOFactory.sol
satyamakgec: I think we have the changes for USDTieredSTO in the 2.1.0 then yes we should bump the version

contracts/modules/STO/USDTieredSTO.sol
satyamakgec: hmmm, As per my knowledge whether the constant is internal or public it always lives in the inline code of the contract. But I will double check for sure.

contracts/proxy/USDTieredSTOProxy.sol
satyamakgec: We don't need to explicitly initialize the `OwnedProxy`. It is similar to `Ownable` contract where there are no arguments in the constructors so it will get automatically initialized. Right now the proxy ownership is given to factory contract which will never able to call any function of the `OwnedProxy` contract. So we could provide the ownership to an issuer by explicitly initializing the `OwnedProxy(SecurityToken)` and change the `_owner()` function definition  to

        function _owner() internal view returns (address) {
                    return IOwnable(SecurityToken).owner();
         }

maxsam4: If I remember correctly, the constructor of the inherited contract is executed automatically only if the inheriting contract does not have a constructor of its own. I'll do a quick test to confirm.

maxsam4: Inherited constructors are also executing automatically. My bad.

maxsam4: It might make sense to just remove the owned part in module proxies for now as we don't support upgrading them (yet)

contracts/modules/STO/ISTO.sol
maxsam4: Just tested, enums and public constants don't affect storage layout.

contracts/modules/STO/USDTieredSTO.sol
satyamakgec: Yes, it lives in the inline code not in the storage so it doesn't affect the storage.
Test gist - 
https://gist.github.com/satyamakgec/ca55f663480828e5eff7c6b9d1ae83b2

contracts/modules/STO/ISTO.sol
satyamakgec: Same for Enums 
Test gist - 
https://gist.github.com/satyamakgec/ca55f663480828e5eff7c6b9d1ae83b2

contracts/proxy/USDTieredSTOProxy.sol
satyamakgec: it only used for getting the implementation contract address, transfer proxy ownership, get the owner contract..... So it doesn't affect anything just the addon. so might make sense to remove the `OwnedProxy` contract and move the `__implementation` variable into the `Proxy` contract.

adamdossa: The reason I created OwnedProxy was to avoid a possible namespace clash with the `implementation` function (by using the `ifOwner` modifier).

adamdossa: In 2.1.0 I can't really see any reason to change the owner to the security token, but in 3.0.0 we'll be re-jigging this stuff to make certain types of upgrade possible (with control split between the factory and issuer). e.g. something like factory can propose an upgrade, issuer actions it.

maxsam4 (approved):

PR #460, opened 2018-12-06
Fetch poly token from pm registry

kostind:

Instead of manually entering the PolyToken on the deployment of each contract, have them fetch it from the PMRegistry.

1. Removed _polyAddress parameter from constructors of all modules and module factories.
2. Removed _polyToken parameter from SecurityTokenRegistry.initialize
3. Removed SecurityTokenRegistry.updatePolyTokenAddress

### DIFF
--- a/contracts/SecurityTokenRegistry.sol
+++ b/contracts/SecurityTokenRegistry.sol
@@ -5,6 +5,7 @@ import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
 import "./interfaces/IOwnable.sol";
 import "./interfaces/ISTFactory.sol";
 import "./interfaces/ISecurityTokenRegistry.sol";
+import "./interfaces/IPolymathRegistry.sol";
 import "./storage/EternalStorage.sol";
 import "./libraries/Util.sol";
 import "./libraries/Encoder.sol";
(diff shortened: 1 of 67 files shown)


maxsam4 (commented):
Update CHANGELOG, please.
Everything else looks good to me.

contracts/SecurityTokenRegistry.sol
```diff
@@ -741,12 +739,11 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
...
-     * @notice Changes the PolyToken address. Only Polymath.
-     * @param _newAddress is the address of the polytoken.
+     * @notice Stores the contract addresses of other key contracts from the PolymathRegistry
      */
-    function updatePolyTokenAddress(address _newAddress) external onlyOwner {
-        require(_newAddress != address(0), "Invalid address");
-        set(POLYTOKEN, _newAddress);
+    function updateFromRegistry() external onlyOwner {
+        address _polymathRegistry = getAddress(Encoder.getKey("polymathRegistry"));
+        set(Encoder.getKey("polyToken"), IPolymathRegistry(_polymathRegistry).getAddress("PolyToken"));
```
satyamakgec: this is an extra step need to perform after calling `initialize` function. As we will never call initialize function again to set the variables because it already presents in the proxy storage. To make thing consistent you can create an internal function of `updateFromRegistry()` that can be called by the `initialize()` function so we can get rid of the extra step of setting up the PolyToken address (If we ever deploy the whole proxy contract again).  
One more thing use constant `POLYTOKEN` variable instead of `Encoder.getKey("polyToken")`

satyamakgec (changes requested):

kostind:
CHANGELOG has been updated.

maxsam4 (approved):

satyamakgec (approved):

PR #467, opened 2018-12-13
Fix rounding off errors

maxsam4:
Fixes (some of) the rounding errors.


maxsam4:
The simulation ran out of DAI. That's new.

adamdossa (approved):

PR #478, opened 2018-12-17
Solidity 0.5.0

maxsam4:
Even after using my refactor tool, there are still a lot of manual changes remaining.
For example, now variables are not available outside their block. So all code like double for loops now have to define vars again. A lot of explicit conversions are no longer allowed directly so they need to change as well. etc.


maxsam4:
Travis is failing cause openzeppelin haven't updated their official package to solidity 0.5.0.

CircleCI's image is debain based instead of ubuntu so it does not have direct access solc binary via ppa. As it's running in docker, it can't use snap either. It needs more config.

Solidity coverage is not working with solidity 0.5.0 so it doesn't matter anyway.

maxsam4:
The project is building fine now (with a custom version of OZ). It will take time to refactor test cases though.

maxsam4:
At 13.5k lines changed and only 10% test cases passing yet.
God bless regex though.

maxsam4:
Both CI are now compiling, testing and generating coverage just fine for the test cases that have been fixed.
Repo is now in a condition to easily run on everyone's system.
To run on your system,
1) `sudo apt install solc` or whatever the mac equivalent is.
2) pull branch `truffle-5` in a new directory
3) `yarn install`
4) `npm run test` to test etc.

maxsam4:
All tests are passing on travis now :)

Tests are failing during coverage though. Probably because solidity-coverage uses web3 0.20 which over rides the newer web3 1.0. Need further inspection.

ps #495 changed 2 files but introduced 23 merge conflicts :/.

kostind:
2_deploy_contracts.js should be re-merged

contracts/FeatureRegistry.sol
```diff
@@ -28,7 +27,7 @@ contract FeatureRegistry is IFeatureRegistry, ReclaimTokens {
      * @param _nameKey is the key for the feature status mapping
      * @param _newStatus is the new feature status
      */
-    function setFeatureStatus(string _nameKey, bool _newStatus) public onlyOwner {
+    function setFeatureStatus(string memory _nameKey, bool _newStatus) public onlyOwner {
```
adamdossa: Any reason this function is public rather than external?

contracts/interfaces/IOwnable.sol
```diff
@@ -10,7 +9,7 @@ interface IOwnable {
     /**
     * @dev Returns owner
     */
-    function owner() external view returns (address);
+    function owner() external view returns(address payable);
```
adamdossa: AFAICT the only time we rely on the owner() being payable is in the Dividend modules. This should probably be improved to have a "refund address" explicitly specified rather than using the owner (which may not be the appropriate address, or unable to receive ETH etc.), and we could then make this non-payable. WDYT?

contracts/interfaces/ISecurityToken.sol
```diff
@@ -24,10 +23,9 @@ interface ISecurityToken {
      * @param _from Sender of transfer
      * @param _to Receiver of transfer
      * @param _value Value of transfer
-     * @param _data Data to indicate validation
      * @return bool
      */
-    function verifyTransfer(address _from, address _to, uint256 _value, bytes _data) external returns (bool);
+    function verifyTransfer(address _from, address _to, uint256 _value) external returns(bool success);
```
adamdossa: How come we lost the `bytes _data` param? (and any reason to name this specific return value)?

contracts/libraries/Util.sol
```diff
@@ -71,12 +71,11 @@ library Util {
      * @param _data Passed data
      * @return bytes4 sig
      */
-    function getSig(bytes _data) internal pure returns (bytes4 sig) {
+    function getSig(bytes memory _data) internal pure returns(bytes4 sig) {
         uint len = _data.length < 4 ? _data.length : 4;
-        for (uint i = 0; i < len; i++) {
-            sig = bytes4(uint(sig) + uint(_data[i]) * (2 ** (8 * (len - 1 - i))));
+        for (uint256 i = 0; i < len; i++) {
+          sig |= bytes4(_data[i] & 0xFF) >> (i * 8);
```
adamdossa, line 77: I'll take your word for it that this is equivalent ;-). I guess the tests would have broken if not.

contracts/modules/Checkpoint/DividendCheckpoint.sol
```diff
@@ -136,16 +132,12 @@ contract DividendCheckpoint is DividendCheckpointStorage, ICheckpoint, Module {
...
-        withPerm(DISTRIBUTE)
-        validDividendIndex(_dividendIndex)
-    {
+    ) public withPerm(DISTRIBUTE) validDividendIndex(_dividendIndex) {
         Dividend storage dividend = dividends[_dividendIndex];
         address[] memory investors = ISecurityToken(securityToken).getInvestors();
         uint256 numberInvestors = Math.min(investors.length, _start.add(_iterations));
         for (uint256 i = _start; i < numberInvestors; i++) {
-            address payee = investors[i];
+            address payable payee = address(uint160(investors[i]));
```
adamdossa, line 147: Can't we do cast directly with `address payable payee = investors[i];`?

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -197,7 +182,7 @@ contract EtherDividendCheckpoint is DividendCheckpoint {
         Dividend storage dividend = dividends[_dividendIndex];
         dividend.reclaimed = true;
         uint256 remainingAmount = dividend.amount.sub(dividend.claimedAmount);
-        address owner = IOwnable(securityToken).owner();
+        address payable owner = address(uint160(IOwnable(securityToken).owner()));
```
adamdossa, line 199: Can we cast directly?

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -211,7 +196,7 @@ contract EtherDividendCheckpoint is DividendCheckpoint {
         Dividend storage dividend = dividends[_dividendIndex];
         uint256 remainingWithheld = dividend.dividendWithheld.sub(dividend.dividendWithheldReclaimed);
         dividend.dividendWithheldReclaimed = dividend.dividendWithheld;
-        address owner = IOwnable(securityToken).owner();
+        address payable owner = IOwnable(securityToken).owner();
```
adamdossa: Probably better to explicitly cast owner to payable here, and not make owner generally payable?

contracts/modules/STO/USDTieredSTOFactory.sol
```diff
@@ -32,17 +34,19 @@ contract USDTieredSTOFactory is ModuleFactory {
...
+    function deploy(bytes calldata _data) external returns(address) {
         address polyToken = _takeFee();
-        address usdTieredSTO = new USDTieredSTOProxy(msg.sender, polyToken, logicContract);
+        address usdTieredSTO = address(new USDTieredSTOProxy(msg.sender, polyToken, logicContract));
         //Checks that _data is valid (not calling anything it shouldn't)
         require(Util.getSig(_data) == IBoot(usdTieredSTO).getInitFunction(), "Invalid data");
+        bool success;
         /*solium-disable-next-line security/no-low-level-calls*/
-        require(address(usdTieredSTO).call(_data), "Unsuccessfull call");
+        (success, ) = address(usdTieredSTO).call(_data);
```
adamdossa: Don't think we need the address cast here.

contracts/tokens/SecurityToken.sol
```diff
@@ -294,27 +286,27 @@ contract SecurityToken is ERC20, ERC20Detailed, ReentrancyGuard, RegistryUpdater
     * @param _module address of module to unarchive
     */
     function removeModule(address _module) external onlyOwner {
```
adamdossa, line 294: Assume this is due to contract size issues?

contracts/interfaces/IOwnable.sol
maxsam4: We can do that but I don't see any harm in keeping it payable. All address pre 0.5 were payable anyway. Another variable/getter will mean slightly bigger size.

contracts/FeatureRegistry.sol
maxsam4: I don't think so. Changed it to `external`

contracts/interfaces/ISecurityToken.sol
maxsam4: Dmitriy probably skipped that by mistake while merging #495 .
There is no reason to name a variable in interfaces apart from clarity nor there is a reason to not name it. It is not part of the bytecode.

contracts/libraries/Util.sol
maxsam4: I got it right after a dozen tries lol. Nobody had manipulated strings like this in solidity 0.5 yet :).

contracts/modules/Checkpoint/DividendCheckpoint.sol
maxsam4: unfortunately not :(. I ranted about this in slack :p.

maxsam4: Unfortunately not. I ranted about this on slack :P

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
maxsam4: We can't directly cast non-payable to payable but owner is already being returned as payable in the interface so we don't need to cast it here. I'll make the change.

If we make owner non-payable in the interface, we'll have to use this casting here.

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
maxsam4: Either way seems fine to me. Casting just looks ugly and takes a few more bytes though. To be honest, I don't even understand why they had to split address into payable and non-payable.

Let me know which way you prefer and I'll change it accordingly.

contracts/modules/STO/USDTieredSTOFactory.sol
maxsam4: fixed.

contracts/tokens/SecurityToken.sol
maxsam4: Oh yes. I totally forgot about this. I also commented out tests that required removal of modules.

Do you prefer to merge this as it as and reduce the size of ST in future or reduce the size before merging?

contracts/SecurityTokenRegistry.sol
```diff
@@ -387,44 +383,41 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
-            set(key, _expiryDate);
+        if (getUint(key) != _expiryDate) set(key, _expiryDate);
         key = Encoder.getKey("registeredTickers_tokenName", _ticker);
-        if (Encoder.getKey(getString(key)) != Encoder.getKey(_tokenName))
-            set(key, _tokenName);
+        if (Encoder.getKey(getString(key)) != Encoder.getKey(_tokenName)) set(key, _tokenName);
         key = Encoder.getKey("registeredTickers_status", _ticker);
-        if (getBool(key) != _status)
-            set(key, _status);
+        if (getBool(key) != _status) set(key, _status);
```
satyamakgec, line 404: Is the one statement in `if` is disallowed by the 0.5?, If not then I believe the earlier was better in reading the understanding.

contracts/SecurityTokenRegistry.sol
```diff
@@ -468,7 +461,12 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
      * @param _tokenDetails is the off-chain details of the token
      * @param _divisible is whether or not the token is divisible
      */
-    function generateSecurityToken(string _name, string _ticker, string _tokenDetails, bool _divisible) external whenNotPausedOrOwner {
+    function generateSecurityToken(
+        string calldata _name,
+        string calldata _ticker,
+        string calldata _tokenDetails,
+        bool _divisible
+    ) external whenNotPausedOrOwner {
```
satyamakgec: I think styling could be improved. 
        
       function generateSecurityToken(
        string calldata _name,
        string calldata _ticker,
        string calldata _tokenDetails,
        bool _divisible
       ) 
            external 
            whenNotPausedOrOwner
       {
        }

In other functions as well. It makes the code more redable.

contracts/libraries/BokkyPooBahsDateTimeLibrary.sol
```diff
@@ -56,18 +55,13 @@ library BokkyPooBahsDateTimeLibrary {
...
         int _month = int(month);
         int _day = int(day);
 
-        int __days = _day
-          - 32075
-          + 1461 * (_year + 4800 + (_month - 14) / 12) / 4
-          + 367 * (_month - 2 - (_month - 14) / 12 * 12) / 12
-          - 3 * ((_year + 4900 + (_month - 14) / 12) / 100) / 4
-          - OFFSET19700101;
+        int __days = _day - 32075 + 1461 * (_year + 4800 + (_month - 14) / 12) / 4 + 367 * (_month - 2 - (_month - 14) / 12 * 12) / 12 - 3 * ((_year + 4900 + (_month - 14) / 12) / 100) / 4 - OFFSET19700101;
```
satyamakgec, line 64: It is a nightmare to understand. can't we go as it was written earlier

satyamakgec (commented):
@maxsam4. In this PR I have a concern related to styling because lot of places we lost redability. (Ex- return, constructor, function declartion statements etc.). 
we could also remove the `updatePolyTokenAddress()` because it can be handled by the Dmitriy changes (`updateFromRegistry()`). It also helps to reduce the size little bit.

maxsam4:
@satyamakgec The styling changes were done by prettier automatically. We can manually change them to our liking in future but I think, for now, we should just let them be.

contracts/SecurityTokenRegistry.sol
maxsam4: To be honest, I prefer the changed style. i.e modifiers in a single line. It's a personal choice though and I am fine with any. I just don't think it's worth it to spend time in refactoring this.

contracts/SecurityTokenRegistry.sol
maxsam4: This has nothing to do with solidity. It's just style preference. Prettier was originally made for Javascript and JS folks prefer to compress if statements like this. We can revert to original one or keep this change. I am fine with any. I have worked with both styles in the past so they both seem kinda natural to me.

contracts/libraries/BokkyPooBahsDateTimeLibrary.sol
maxsam4: We aren't even supposed to change this file. Once bokky updates the library for solidity 0.5, we'll just use his :).

satyamakgec:
hmm... It is not a big task to tackle so we can merge this PR with all the changes. 
@comeonbuddy could you please take care of styling in this PR. I will hardly take 2 hours to work it manually.

contracts/tokens/SecurityToken.sol
adamdossa: Let's merge and create an issue to uncomment the function as a reminder.

adamdossa (approved):

Issue #484, opened 2018-12-19
Eternal Storage missing public functions


maxsam4:
In the [eternal storage contract](https://github.com/PolymathNetwork/polymath-core/blob/master/contracts/storage/EternalStorage.sol), There are public functions for getting Elementary Data Type variables like uint but the functions to get Arrays are marked internal. Those functions can also be useful to outside world and should be marked public.

functions like
```
function getArrayAddress(bytes32 _key) internal view returns(address[]) {
        return addressArrayStorage[_key];
}
``` 
should become
```
function getArrayAddress(bytes32 _key) public view returns(address[]) {
        return addressArrayStorage[_key];
}
```

dev1644:
I am working on it.

maxsam4:
Closed by https://github.com/PolymathNetwork/polymath-core/pull/488
Thank you @dev1644 for your contribution.

Issue #483, opened 2018-12-19
Eternal storage redundant functions


maxsam4:
There are a few redundant functions in the [eternal storage contract](https://github.com/PolymathNetwork/polymath-core/blob/master/contracts/storage/EternalStorage.sol)
like
```
function getBool(bytes32 _key) internal view returns (bool) {
        return boolStorage[_key];
}
```
and
```
function getBoolValues(bytes32 _variable) public view returns(bool) {
        return boolStorage[_variable];
}
```
The internal functions can be removed and code can be refactored to use the public function instead.

dev1644:
I am working on it.

maxsam4:
Closed by https://github.com/PolymathNetwork/polymath-core/pull/488
Thank you @dev1644 for your contribution

PR #488, opened 2018-12-20
483 484 fix

dev1644:
### Please check if the PR fulfils these requirements
- [x] Tests for the changes have been added (for bug fixes/features)

Removed the Redundant functions from eternal storage contract as mentioned in Issue #483, Converted Internal functions to public ones as mentioned in Issue #484 & there was a function named getArrayString, I have removed that function because it was returning Array of String, which is not possible as it isn't supported by Current ABIENCODER."

Removed the redundant functions and renamed newly added functions.

Issue #483 
Issue #484 

### What is the new behavior?


contracts/storage/EternalStorage.sol
```diff
@@ -232,9 +195,16 @@ contract EternalStorage {
...
+    /// @notice Get function use to get the value of the singleton state variables
+    /// Ex1- string public version = "0.0.1";
+    /// string _version = getString(keccak256(abi.encodePacked("version"));
+    /// Ex2 - assert(temp1 == temp2); replace to
+    /// assert(getUint(keccak256(abi.encodePacked(temp1)) == getUint(keccak256(abi.encodePacked(temp2));
+    /// Ex3 - mapping(string => SymbolDetails) registeredSymbols; where SymbolDetails is the structure having different type of values as
+    /// {uint256 date, string name, address owner} etc.
+    /// string _name = getString(keccak256(abi.encodePacked("registeredSymbols_name", "TOKEN"));
 
     function getUintValues(bytes32 _variable) public view returns(uint256) {
```
maxsam4: Can you please rename functions like `getUintValues` to `getUintValue` as they are returning only 1 value.
Also, update the code comment above to show the new name.

maxsam4 (changes requested):

contracts/storage/EternalStorage.sol
dev1644: Done.

maxsam4 (approved):

adamdossa:
Thanks @dev1644 - changes look good.

The only issue is that `getArrayString` can't be public until we can use the ABIencoderV2 in production. I'll add this as an issue to remind us to re-add this in the future when ABIencoderv2 is production ready.

PR #494, opened 2018-12-21
Get all created STs

kostind:

Added `getTokens` method to SecurityTokenRegistry.


adamdossa (changes requested):
Hey @kostind - I think this approach isn't 100% optimal. It is possible for a security token to have address(0) as it's actual owner (i.e. if the issuer calls `renounceOwnership` on the token) and this approach means that it won't be possible to query for tokens where the actual owner is address(0). Could you have a go at refactoring without using address(0) as a sentinel for all addresses?

kostind:
@adamdossa done

adamdossa (approved):

PR #499, opened 2019-01-02
Optimize bytecode str

satyamakgec:
new size of the STR 20.79 KB


maxsam4:
I thought we were gonna proxy all the getter requests through the main contract to the other contract.
It will take ~1Kb more but will remove the hassle of keeping two contract addresses. what do you think?

satyamakgec:
we don't need to maintain the two address from the dapp side only need to use the ABI.

maxsam4:
Yeah, My bad. I didn't notice you inherited the proxy contract in STR :).

contracts/SecurityTokenRegistry.sol
```diff
@@ -203,6 +207,19 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
         set(POLYTOKEN, IPolymathRegistry(polymathRegistry).getAddress("PolyToken"));
     }
 
+    /**
+     * @notice Set the getter contract address
+     * @param _getterContract Address of the contract  
+     */
+    function setGetterRegistry(address _getterContract) public onlyOwner {
+        require(_getterContract != address(0));
+        set(Encoder.getKey("STRGetter"), _getterContract);
```
maxsam4: We can hardcode the STRGetter key.

maxsam4 (commented):
Looks good. Just one possible optimization.

contracts/SecurityTokenRegistry.sol
satyamakgec: Yes, we can hardcode. But It will restrict us to redeploy the STR as well if there is any change in the getter functions or we need to add more getters in the `STRGetter` contract

maxsam4: Why? I am asking to hardcode just the key `Encoder.getKey("STRGetter")`
Like we have hardcoded other keys.
Just to save the computation required to calculate the key.

satyamakgec: Ohh My bad, I was thinking you are talking about the address.

maxsam4 (approved):

PR #513, opened 2019-01-08
add getRestrictedData() function in VRTM

satyamakgec:


contracts/modules/TransferManager/VolumeRestrictionTM.sol
```diff
@@ -449,12 +461,28 @@ contract VolumeRestrictionTM is VolumeRestrictionTMStorage, ITransferManager {
...
         require(individualRestriction[_holder].endTime != 0, "Not present");
         individualRestriction[_holder] = VolumeRestriction(0, 0, 0, 0, RestrictionType(0));
+        _deleteHolderFromList(_holder, uint8(TypeOfPeriod.OneDay));
         userToBucket[_holder].lastTradedDayTime = 0;
         userToBucket[_holder].sumOfLastPeriod = 0;
         userToBucket[_holder].daysCovered = 0;
         emit IndividualRestrictionRemoved(_holder);
     }
 
+    function _deleteHolderFromList(address _holder, uint8 _typeOfPeriod) internal {
```
maxsam4: Add comment code. specifically, mention that `_typeOfPeriod` is what should NOT be deleted.

contracts/modules/TransferManager/VolumeRestrictionTMStorage.sol
```diff
@@ -25,6 +27,15 @@ contract VolumeRestrictionTMStorage {
         uint256 dailyLastTradedDayTime;
     }
 
+    struct RestrictedHolder {
+        // 1 represent true & 0 for false
+        uint8 seen;
+        // Type of period will be enum index of TypeOfPeriod enum
+        uint8 typeOfPeriod;
+        // Index of the array where the holder address lives
+        uint64 index;
```
maxsam4: We can use `uint128` instead of `uint64`

maxsam4 (changes requested):

contracts/modules/TransferManager/VolumeRestrictionTMStorage.sol
satyamakgec: Yes, we could But 64 also work for us. For filling the remaining space we could use 128

maxsam4: Let's just go with uint128 because we have space. No need to limit ourself where we don't need to :)

maxsam4 (approved):

satyamakgec:
@VictorVicente please take a look at the `getRestrictedAddress()` function return values and confirm that you need the same data. I will remove the earlier one once you confirm that new one is according to the requirement.

satyamakgec:
@maxsam4 could you please review the latest changes.

maxsam4 (approved):

contracts/modules/TransferManager/VolumeRestrictionTM.sol
```diff
@@ -143,8 +143,8 @@ contract VolumeRestrictionTM is VolumeRestrictionTMStorage, ITransferManager {
             exemptIndex[_wallet] = exemptAddresses.length;
             exemptList[_wallet] = true;
         } else {
-            exemptAddresses[exemptIndex[_wallet]] = exemptAddresses[exemptAddresses.length -1];
-            exemptIndex[exemptAddresses[exemptIndex[_wallet]]] = exemptIndex[_wallet];
+            exemptAddresses[exemptIndex[_wallet] - 1] = exemptAddresses[exemptAddresses.length -1];
```
adamdossa: How come we do this rather than using:
`exemptIndex[_wallet] = exemptAddresses.length - 1;`
above? (so the indexes aren't off by one)?

maxsam4: Indexes are off by one always. While adding exemptAddresses, the address is pushed first and then the length of the array is used as the index.

Basically, the index we are using starts from 1 instead of 0. It's more of position in the array rather than the index.

PR #515, opened 2019-01-08
[CLI] Cli text and improvements

VictorVicente:

Fixes and updates for the CLI

No


CLI/commands/transfer_manager.js
```diff
@@ -396,8 +396,8 @@ async function generalTransferManager() {
       let now = Math.floor(Date.now() / 1000);
       let fromTime = readlineSync.questionInt(`Enter the time (Unix Epoch time) when the sale lockup period ends and the investor can freely sell his tokens (now = ${now}): `, { defaultInput: now });
       let toTime = readlineSync.questionInt(`Enter the time (Unix Epoch time) when the purchase lockup period ends and the investor can freely purchase tokens from others (now = ${now}): `, { defaultInput: now });
-      let oneHourFromNow = Math.floor(Date.now() / 1000 + 3600);
-      let expiryTime = readlineSync.questionInt(`Enter the time till investors KYC will be validated (after that investor need to do re - KYC) (1 hour from now = ${oneHourFromNow}): `, { defaultInput: oneHourFromNow });
+      let oneYearFromNow = Math.floor(Date.now() / 1000 + (60 * 60 * 24 * 365));
+      let expiryTime = readlineSync.questionInt(`Enter the time until the investors KYC will be valid (after this time expires, the investor must re-do KYC) (1 year from now = ${oneYearFromNow}): `, { defaultInput: oneYearFromNow });
       let canBuyFromSTO = readlineSync.keyInYNStrict('Is the investor a restricted investor?');
```
F-OBrien: This question needs to be changed as it causes confusion. It had me scratching my head a few weeks ago when I was getting invalid transfers. Similarly it needs to be changed in the Token manager.
```suggestion
       let canBuyFromSTO = readlineSync.keyInYNStrict('Can the investor buy from security token offerings?');
```

VictorVicente: Thanks @F-OBrien. I always try to keep these texts as similar to comments in contracts bun in this case this is much descriptive than the original.

F-OBrien: No Problem. This also needs to be changed in the Token Manager. There are a few other things that can be fixed in the CLI's also. Take a look at some of the commits #541 I made against Dev-3.0.0  as some of the changes should be applied to Dev-2.1.0 also. (I see Adam has requested you review it earlier today)

VictorVicente: Yes, I checked them. A couple of them were already included in this PR. Some others are changes regarding 3.0.0 version of contracts, so I can't include them here.
Once dev-2.1.0 is merged into dev-3.0.0, I'll approve and merge your PR too.

CLI/commands/investor_portal.js
```diff
@@ -498,8 +498,8 @@ async function investCappedSTO(currency, amount) {
     if (raiseTypes[0] == 'POLY') {
         let userBalance = await polyBalance(User.address);
         if (parseInt(userBalance) >= parseInt(cost)) {
-            let allowance = await polyToken.methods.allowance(STOAddress, User.address).call();
-            if (allowance < costWei) {
+            let allowance = await polyToken.methods.allowance(User.address, STOAddress).call();
+            if (parseInt(allowance) < parseInt(costWei)) {
```
F-OBrien, line 502: This is a question rather than a change request, again it's something that caught me out when I was working on a CLI for a STO contract I'm creating. In multiple CLI files there were less than or greater than comparators of number strings. In my CLI's I did like you've done and parsed them to integers or float as a quick fix but I was wondering if some of them at least should be compared as BN instead of integers/float?

VictorVicente: You are right, most of these comparisons should be done by BN library instead of parse. But as you said, it's quicker this way and I'm planning isolate this type of thins to an utils library to avoid repeat it everywhere.
Unfortunately I didn't have the time to do it but I will soon

CPSTL (approved):

PR #521, opened 2019-01-14
dev 2.1 > dev 3.0

adamdossa:


satyamakgec:
@adamdossa @kostind if you are getting short handed then please let me know. I will help you out. As I need dev-2.1.0 changes in dev-3.0.0 to solve the size issue of the ST and need new modules.

satyamakgec:
@adamdossa @kostind if you are getting short handed then please let me know. I will help you out. As I need dev-2.1.0 changes in dev-3.0.0 to solve the size issue of the ST and need new modules.

contracts/storage/VolumeRestrictionTMStorage.sol
```diff
@@ -45,10 +45,15 @@ contract VolumeRestrictionTMStorage {
...
-    mapping(address => VolumeRestriction) public individualDailyRestriction;
+
+    struct Restrictions {
+        // Restriction stored corresponds to a particular token holder
+        mapping(address => VolumeRestriction) individualRestriction;
+        // Daily restriction stored corresponds to a particular token holder
+        mapping(address => VolumeRestriction) individualDailyRestriction;
+    }
+
+    Restrictions restrictions;
```
maxsam4: Does this actually save some size?

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -683,61 +626,50 @@ contract USDTieredSTO is USDTieredSTOStorage, STO, ReentrancyGuard {
...
      * @return uint256 Total number of tokens minted
      */
-    function getTokensMinted() public view returns(uint256) {
-        uint256 tokensMinted;
+    function getTokensMinted() public view returns (uint256 tokensMinted) {
         for (uint256 i = 0; i < tiers.length; i++) {
             tokensMinted = tokensMinted.add(tiers[i].mintedTotal);
         }
-        return tokensMinted;
     }
```
F-OBrien, line 651: Optimization Suggestion?
I could be wrong here but is there a logic to using `tokensMinted` instead of `totalTokensSold`? Are they not essentially the same thing? `totalTokensSold` only seems to be update after the STO is finaliszed. Instead there currently is an additional function that regularly calculates `tokensMinted`.  If `totalTokensSold` was update as tokens were minted instead it would save the need for `getTokensMinted()`, which is called multiple times and recalculates each time. It could simply be replaced with `totalTokensSold`. This would then allow you to remove `getTokensMinted()` and also remove the `getTokensSold()` function as `totalTokensSold` already has a public getter.
So by adding `totalTokensSold = totalTokensSold.add(tierPurchasedTokens)` after `tierData.mintedTotal = tierData.mintedTotal.add(tierPurchasedTokens)` in the `_calculateTier` function I think it allows this optimization.
The `finalize()` function could then also be tidied up a bit.

adamdossa: Thanks @F-OBrien - I've marked this as an issue to see where it can be cleaned up. The difference is meant to be the amount of tokens which were actually sold through the STO, vs. the number of tokens which have been minted (including minting back to the reserve wallet). As you say we could probably clean this up though. #535

contracts/storage/VolumeRestrictionTMStorage.sol
adamdossa: Not really, but does tidy things up a bit.

satyamakgec (approved):
Everything is good.

PR #522, opened 2019-01-14
Datastore

maxsam4:
This PR contains the basic data store and a sample module.
Existing modules will be refactored to use data store in individual PRs.


maxsam4:
Wait. Why does this PR have so many commits :/
I guess github didn't detect the merge.

Anyway, it's WIP, I'll rebase it later.

contracts/datastore/DataStore.sol
```diff
@@ -0,0 +1,362 @@
...
+        );
+        _;
+    }
+
+    modifier validKey(bytes32 _key) {
+        require(_key != bytes32(0), "Missing key");
+        _;
+    }
+
+    modifier validInputLength(uint256 _keyLength, uint256 _dataLength) {
```
satyamakgec: Name could be changed to `validArrayLength()`.  As both parameters are uint256 it creates confusion as it gets clubbed with the name `validInputLength`.

contracts/modules/Experimental/TransferManager/KYCTransferManager.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    function verifyTransfer(address /*_from*/, address _to, uint256 /*_amount*/, bytes memory /* _data */, bool /* _isTransfer */) 
+        public 
+        returns (Result) 
+    {
+        if (!paused) {
+            bytes32 key = _getKYCKey(_to);
+            IDataStore dataStore = IDataStore(ISecurityToken(securityToken).dataStore());
+            if (dataStore.getUint256(key) > 0)
+                return Result.VALID;
```
satyamakgec: `checkKYC` can be reused here.

contracts/modules/Experimental/TransferManager/KYCTransferManager.sol
```diff
@@ -0,0 +1,102 @@
...
+        }
+        return Result.NA;
+    }
+
+    function modifyKYC( address _investor, bool _kycStatus) public withPerm(KYC_PROVIDER) {
+        _modifyKYC(_investor, _kycStatus);
+    }
+
+    function _modifyKYC(address _investor, bool _kycStatus) internal {
+        IDataStore dataStore = IDataStore(ISecurityToken(securityToken).dataStore());
```
satyamakgec, line 54: we could make a getter function to return a datastore module address in the module.sol

satyamakgec (commented):
We could use some shorter names for the external functions of DataStore contract. It will make ease to write the data from the other modules. (It will make our code neat and readable).
other than this everything is good.

contracts/modules/Experimental/TransferManager/KYCTransferManager.sol
maxsam4: Yeah, I didn't do that in this PR because I am split if we should make a getter function or just store the data store address in the constructor.
Using the getter will mean higher gas but every time the latest data store will be fetched. Storing it as a variable will mean lower gas but we'll need to update ds address in each module if we change ds.

data store will rarely (if ever) be changed after deployment but storing the data store in every module may end up being more expensive if/when rent is introduced.

contracts/modules/Experimental/TransferManager/KYCTransferManager.sol
maxsam4: Yep, will do that.

contracts/datastore/DataStore.sol
maxsam4: I actually had it named `validArrayLength` earlier but changed it to `validInputLength` because it seemed better to me.
Anyway, I don't really have a strong preference for any. I'll change it back to `validArrayLength`.

contracts/datastore/DataStore.sol
```diff
@@ -0,0 +1,362 @@
...
+        _;
+    }
+
+    /**
+     * @dev Changes security token atatched to this data store
+     * @param _securityToken address of the security token
+     */
+    function setSecurityToken(address _securityToken) external onlyOwner {
+        if(address(securityToken) != address(0)) {
+            require(msg.sender == IOwnable(address(securityToken)).owner(), "Unauthorized");
```
adamdossa: Isn't this already done through the `onlyOwner` modifier?

adamdossa: Will modify to just have the modifier (which is needed to stop someone changing the security token to address(0) which would break things).

maxsam4: `onlyOwner` wasn't supposed to be added to that function. We don't even need `onlyOwner`.

My original thinking was that first setSecurityToken call must be free to allow initialization. After the initialization, onlyOwner should kick in. 

I don't exactly remember why I added onlyOwner. Probably just a reflex 😂.

maxsam4: I decided to go with proxy deployment and Initialized st address in the proxy so I never had to call the set function before initialization.

adamdossa: Thinking a bit more about this, this seems a slightly dangerous operation. An issuer could assign the security token address to be anything (i.e. not an actual security token) which means that the data could then be updated adhoc. The original ST though would still think the data store belonged to it.

I do think it is useful to be able to assign a dataStore to a new token (so that if an issuer wants to migrate its data to a new token it can do) but wondering whether we should also:
  - 1. check the new address is a ST (using the STR)
  - 2. tell the original underlying token that it no longer owns the dataStore
  - 3. not allow address(0) to be used as the new ST address

maxsam4: If we do 1), we won't need 3). If gas limit allows, I think we should add the check just to minimise mistaken address changes.

2) I don't think we should prevent or worry too much about this. Issuer has a lot more power anyway. In case of upgrading token, Keeping the data store active for both tokens (read for old, write for new) can be useful to create a bridge that will help people migrate their security tokens.

adamdossa: OK - for now I've just made it so that the function has an onlyOwner modifier. As you say, for initialisation it is done through the constructor, so any subsequent changes should only be made by the ST owner. I'll add checking in the STR as a subsequent PR - not too worried about gas costs as it will be a rare operation.

adamdossa (approved):

contracts/datastore/DataStore.sol
maxsam4: I meant to say code size limit and not gas limit :)

PR #524, opened 2019-01-15
EIP-1594/1644/1643/1410 implementation

satyamakgec:
TODO:

- [x] Test suite 

- [ ] Documentation

- [ ] Implementation of ERC1410


contracts/interfaces/ITransferManager.sol
```diff
@@ -13,4 +13,8 @@ interface ITransferManager {
         TransferManagerEnums.Result
```
adamdossa: I wonder whether we should having something like:
`executeTransfer` (non-view) and `verifyTransfer` (view) rather than overloading `verifyTransfer`

contracts/libraries/TokenLib.sol
```diff
@@ -278,4 +280,58 @@ library TokenLib {
...
+            bool isInvalid = false;
+            bool isValid = false;
+            bool isForceValid = false;
+            // Use the local variables to avoid the stack too deep error
+            transfersFrozen = false; // bool unarchived = false;
+            byte reasonCode;
+            for (uint256 i = 0; i < modules.length; i++) {
+                if (!modulesToData[modules[i]].isArchived) {
+                    transfersFrozen = true;
+                    (TransferManagerEnums.Result valid, byte reason) = ITransferManager(modules[i]).verifyTransfer(from, to, value, data);
```
adamdossa: Don't we need to call the view / non-view version of the function here depending on whether a transfer is actually happening?

adamdossa (approved):

PR #526, opened 2019-01-15
Dividend improvements

adamdossa:
- Added `_wallet` as a configuration parameter
  - Added `changeWallet` to allow `wallet` to be modified
  - Added Claim & Tax (actual & estimates) to `getDividendsProgress`. If `claimed` is true, amounts are actual, if `claimed` is false amounts are estimates. `resultAmount` is now the claimed amount minus any withholding tax.
```
    /**
     * @notice Retrieves list of investors, their claim status and whether they are excluded
     * @param _dividendIndex Dividend to withdraw from
     * @return address[] list of investors
     * @return bool[] whether investor has claimed
     * @return bool[] whether investor is excluded
     * @return uint256[] amount of withheld tax (estimate if not claimed)
     * @return uint256[] amount of claim (estimate if not claimeed)
     * @return uint256[] investor balance
     */
```

### DIFF
--- a/CLI/commands/dividends_manager.js
+++ b/CLI/commands/dividends_manager.js
@@ -123,14 +123,16 @@ async function configExistingModules(dividendModules) {
 async function dividendsManager() {
   console.log(chalk.blue(`Dividends module at ${currentDividendsModule.options.address}`), '\n');
 
+  let wallet = await currentDividendsModule.methods.wallet().call();
   let currentDividends = await getDividends();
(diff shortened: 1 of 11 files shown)


satyamakgec (approved):

maxsam4 (approved):

maxsam4:
Changes look good to me. 
One small thing that was a bit unusual was that setWallet event also contains address of old wallet. I don't think we need the old wallet address there.

PR #530, opened 2019-01-18
add the 2.1.0 changes

satyamakgec:


maxsam4 (commented):
Changelog seems comprehensive enough to me.
One thing we should probably change is that we should rename v1.5.0 to v2.0.0

adamdossa (approved):

PR #533, opened 2019-01-24
Signed Transfer Manager

maxsam4:
This PR includes STM and proper test case with signed data.
The module uses datastore to store its data.


contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,207 @@
...
+            uint256 nonce;
+            uint256 expiry;
+            bytes memory signature;
+            (targetAddress, nonce, expiry, signature) = abi.decode(_data, (address, uint256, uint256, bytes));
+
+            if (address(this) != targetAddress || signature.length == 0 || _checkSignatureIsInvalid(signature) || expiry < now)
+                return Result.NA;
+
+            bytes32 hash = keccak256(abi.encodePacked(targetAddress, nonce, expiry, _from, _to, _amount));
+            bytes32 prependedHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
```
satyamakgec: Encoder Library could be used

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,207 @@
...
+            address signer = _recoverSignerAdd(prependedHash, signature);
+
+            if (!_checkSigner(signer)) {
+                return Result.NA;
+            } else if(_isTransfer) {
+                _invalidateSignature(signature);
```
satyamakgec, line 118: instead of `SignatureInvalidated` event, we could emit `SignatureUsed` or something +ve event as transaction get succeed. For  `invalidateSignature()` function this event makes sense but for a successful transfer, it will create confusion.

contracts/modules/Experimental/TransferManager/SignedTransferManagerFactory.sol
```diff
@@ -0,0 +1,109 @@
...
+    function getInstructions() external view returns(string memory) {
+        return "Allows an issuer to maintain a list of signers who can validate transfer request using signatures. A mapping is used to track valid signers which can be managed by the issuer. verifytransfer function takes in a signature and if the signature is valid, it will verify the transfer. invalidSigature function allow the signer to make a signature invalid after it is signed. Init function takes no parameters.";
+    }
+
+    /**
+     * @notice Get the tags related to the module factory
+     */
+    function getTags() public view returns(bytes32[] memory) {
+        bytes32[] memory availableTags = new bytes32[](2);
+        availableTags[0] = "General";
```
satyamakgec: Should be "Signed"

contracts/modules/Experimental/TransferManager/SignedTransferManagerFactory.sol
```diff
@@ -0,0 +1,109 @@
...
+    //         require(polyToken.transferFrom(msg.sender, owner, setupCost), "Failed transferFrom because of sufficent Allowance is not provided");
+    //     address signedTransferManager = new SignedTransferManager(msg.sender, address(polyToken));
+    //     emit GenerateModuleFromFactory(address(signedTransferManager), getName(), address(this), msg.sender, setupCost, now);
+    //     return address(signedTransferManager);
+    // }
+
+     function deploy(bytes calldata /* _data */) external returns(address) {
+        address polyToken = _takeFee();
+        SignedTransferManager signedTransferManager = new SignedTransferManager(msg.sender, polyToken);
+        emit GenerateModuleFromFactory(address(signedTransferManager), getName(), address(this), msg.sender, now);
```
satyamakgec: Isn't it follow the proxy approach of deployment ?

satyamakgec (changes requested):

contracts/modules/Experimental/TransferManager/SignedTransferManagerFactory.sol
adamdossa: For "Experimental" modules we haven't been using the Proxy deployment approach.

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
adamdossa: I'll refactor a bit to use this (and in other places if needed)

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+    * @notice function to remove or add signer(s) onto the signer mapping
+    * @param _signers address array of signers
+    * @param _signersStats bool array of signers stats
+    */
+    function updateSigners(address[] calldata _signers, bool[] calldata _signersStats) external withPerm(ADMIN) {
+        require(_signers.length == _signersStats.length, "Array length mismatch");
+        IDataStore dataStore = IDataStore(ISecurityToken(securityToken).dataStore());
+        for(uint256 i=0; i<_signers.length; i++) {
```
adamdossa, line 78: Interesting to store Signer addresses in the data store. I guess this makes sense, esp. if "signers" are going to be used in other places and would make upgrading the module easier (via a remove and re-add).

PR #534, opened 2019-01-25
usdtsto optimization

maxsam4:
USDTieredSTO -  23.9375 KB

### DIFF
--- a/contracts/modules/STO/USDTieredSTO.sol
+++ b/contracts/modules/STO/USDTieredSTO.sol
@@ -607,10 +607,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
             return IOracle(_getOracle(bytes32("POLY"), bytes32("USD"))).getPrice();
         } else if (_fundRaiseType == FundRaiseType.SC) {
             return 10**18;
-        } //else { it's a view function so reverting and returning 0 rate is same for outside EVM world. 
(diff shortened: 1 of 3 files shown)


contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -683,7 +670,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO, ReentrancyGuard {
      * @return uint256[] array of minted tokens in each fund raise type
      */
     function getTokensMintedByTier(uint256 _tier) public view returns(uint256[] memory) {
-        require(_tier < tiers.length, "Invalid tier");
+        //require(_tier < tiers.length, "Invalid tier"); It will revert due to array out of bounds anyway
```
satyamakgec: I am ok for this. only need to confirm that for any internal call of this function make parent function revert or not as it will lead the `Invalid opcode` error.

satyamakgec (approved):

contracts/modules/STO/USDTieredSTO.sol
maxsam4: `Invalid opcode` is revert as well. The only difference is that the error message will say "Invalid opcode" rather than "Invalid tier". It can make debugging a tiny bit more confusing if we at some point hit this error but the chances of us facing an error where we are going over tier limit are next to none.

PR #538, opened 2019-02-01
Migrate data of GTM

satyamakgec:


contracts/libraries/VersionUtils.sol
```diff
@@ -106,4 +106,25 @@ library VersionUtils {
         return _unpackVersion;
     }
 
+
+    /**
+     * @notice Used to packed the KYC data
+     */
+    function packKYC(uint64 _a, uint64 _b, uint64 _c, uint8 _d, uint8 _e, uint8 _f) internal pure returns(uint256) {
+        return (uint256(_a) << 184) | (uint256(_b) << 120) | (uint256(_c) << 56) | (uint256(_d) << 24) | (uint256(_e) << 16) | uint256(_f);
```
maxsam4: The data isn't tightly packed here. We only need 8 bits for uint8. Currently, there is space wasted between e-f and c-d.

contracts/libraries/VersionUtils.sol
```diff
@@ -106,4 +106,25 @@ library VersionUtils {
...
+    /**
+     * @notice Used to convert packed data into KYC data
+     * @param _packedVersion Packed data
+     */
+    function unpackKYC(uint256 _packedVersion) internal pure returns(uint64 fromTime, uint64 toTime, uint64 expiryTime, uint8 canBuy, uint8 added, uint8 accredited) {
+        fromTime = uint64(_packedVersion >> 184);
+        toTime = uint64(_packedVersion >> 120);
+        expiryTime = uint64(_packedVersion >> 56);
+        canBuy = uint8(_packedVersion >> 24);
+        added = uint8(_packedVersion >> 16);
```
maxsam4: These will have to change according to the changes in packKYC

maxsam4 (changes requested):
Only partial review completed yet.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -211,18 +226,23 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
-        if (_canBuyFromSTO) {
-            canBuyFromSTO = 1;
-        }
-        if (whitelist[_investor].added == uint8(0)) {
-            investors.push(_investor);
+        uint8 added;
+        uint8 canBuyFromSTO;
+        uint8 accredited;
+        IDataStore dataStore = IDataStore(getDataStore());
+        (,,,,added,) = _getValues(_investor, dataStore);
```
maxsam4: For gas optimization, we can create a dedicated function for this which just bit shift added and return only added.

contracts/modules/TransferManager/GeneralTransferManager.sol
```diff
@@ -300,10 +331,10 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
     /**
      * @notice Internal function used to check whether the investor is in the whitelist or not
             & also checks whether the KYC of investor get expired or not
-     * @param _investor Address of the investor
+     * @param _expiryTime Expiry time of the investor
      */
-    function _onWhitelist(address _investor) internal view returns(bool) {
-        return (whitelist[_investor].expiryTime >= uint64(now)); /*solium-disable-line security/no-block-members*/
+    function _onWhitelist(uint64 _expiryTime) internal view returns(bool) {
+        return (_expiryTime >= uint64(now)); /*solium-disable-line security/no-block-members*/
```
maxsam4, line 336: Do we need this function anymore? it's just `_expiryTime >= uint64(now)`. Calling and defining a function takes gas.

maxsam4 (commented):
LGTM. Left some minor comments regarding optimization.

contracts/modules/TransferManager/GeneralTransferManager.sol
maxsam4: At the very least, we should rename this function.

adamdossa: Aren't all internal functions just in-lined anyway? Having this function (and corresponding _validLockTime) makes the code marginally more readable IMO.

adamdossa:
I modified a couple of the function names and added a _validLockTime function to make code a bit more readable.

contracts/modules/TransferManager/GeneralTransferManager.sol
maxsam4: internal functions of libraries are inlined but normal internal functions of a contract are not inlined. they are treated like normal functions.

The change won't be saving much gas anyway. we can keep the function.

adamdossa: A small test in Remix shows a very small gas difference ~50 gas between using an internal function and manually in-lining the code. Given the tiny difference I think readability wins here.

adamdossa:
I also wonder if we could do something more sensible with the `investors` data. We currently have a list of investors (append-only) in the token, and the list of `investors` (KYC data individuals) in the GTM. We could think about consolidating this in the dataStore possibly in a single data set. This could be a separate PR though.

maxsam4 (approved):

PR #539, opened 2019-02-01
Use USD in place of POLY

adamdossa:
Use USD pricing instead of POLY for ticker registration, token generation and module costs.

All prices are now denominated in USD, and converted to POLY at the point in time the transaction is executed.

Possible concern is calculating the correct amount of POLY to approve / transfer ahead of the operation, but this could be mitigated by a more stable oracle.


contracts/SecurityTokenRegistry.sol
```diff
@@ -234,9 +253,7 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
         require(_owner != address(0), "Owner should not be 0x");
         require(bytes(_ticker).length > 0 && bytes(_ticker).length <= 10, "Ticker length range (0,10]");
         // Attempt to charge the reg fee if it is > 0 POLY
-        uint256 tickerFee = getUintValue(TICKERREGFEE);
```
maxsam4, line 237: `TICKERREGFEE` is now fees in USD, I suppose? All comments still say that it is the fee in Poly. Can you please update the comments if it's USD indeed.

maxsam4 (changes requested):
Changes look good to me. Just some code comments need updating.

contracts/SecurityTokenRegistry.sol
maxsam4: Same goes for `STLAUNCHFEE` / `_stLaunchFee`

adamdossa: Thanks - I went through the code and updated all the comments I could find.

maxsam4 (approved):

satyamakgec (approved):
I already reviewed the PR but forgot to Approved....

PR #543, opened 2019-02-05
Remove old test file

maxsam4:
Deleted test/s_v130_to_v140_upgrade.js


F-OBrien:
If this file is deleted reference to it can also be removed from scripts/test.sh

adamdossa (approved):

PR #545, opened 2019-02-07
Auto deploy to Kovan

maxsam4:
Every time a merge is made to master/dev-2.1.0/dev-2.2.0/dev-3.0.0, A fresh set of polymath core contracts will be deployed on Kovan.

Do NOT use these in production. These are only to ease internal testing. To get the address. open the details of check named `auto-deploy` ran by CircleCI and then expand the output of `node_modules/.bin/truffle deploy --network kovan` command.

It costs about 0.25 eth to deploy one set of polymath core. I have topped off the deployer with 20 keth so we should be good for 80 merges to master/dev branches i.e. 80 pull request merges. I'll top it off again when needed.

ps the commit `Use local truffle` has the auto-deploy check finished successfully and the output can be used. The next commit edits the config so that deployment is done only on merge to master/dev branches.

EDIT: A notification with commit details and PM registry address will be posted in slack for easy access.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,6 @@ workflows:
   commit:
     jobs:
       - coverage
-      - deploy_kovan
   daily-builds:
     triggers:
       - schedule:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,9 +68,7 @@ jobs:
       - run: node --version
       - run: truffle version
       - run: mv truffle-ci.js truffle-config.js
-      - run: 
-          command: node_modules/.bin/truffle deploy --network kovan
-          no_output_timeout: 1h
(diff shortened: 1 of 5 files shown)


satyamakgec:
It seems good to me as it is working. I need to understand the reason behind this change why we are adding this step into CI? @maxsam4

maxsam4:
> It seems good to me as it is working. I need to understand the reason behind this change why we are adding this step into CI? @maxsam4

This is just to make it easier to test for dApp/CLI folks. Already deployed contracts will mean they don't have to waste time in deploying it on their own. Also, everyone will now have same contracts to test against so it will be easier to keep trace when debugging.

satyamakgec (approved):

PR #546, opened 2019-02-07
Stable oracle

adamdossa:
Adds a "StableOracle" which only updates it's price if the latestPrice is more than x% (configurable) from the last served price. This allows USD > POLY conversions which are less likely to change and an improved UX for this feature.


maxsam4 (approved):
LGTM

satyamakgec (approved):

PR #547, opened 2019-02-07
add getTokensByDelegate

adamdossa:
add getTokensByDelegate:
```
    /**
     * @notice Returns the list of tokens to which the delegate has some access
     * @param _delegate is the address for the delegate
     * @dev Intention is that this is called off-chain so block gas limit is not relevant
     */
    function getTokensByDelegate(address _delegate) external view returns(address[]);
```

### DIFF
--- a/contracts/SecurityTokenRegistry.sol
+++ b/contracts/SecurityTokenRegistry.sol
@@ -4,7 +4,9 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "./interfaces/IOwnable.sol";
 import "./interfaces/ISTFactory.sol";
 import "./interfaces/IERC20.sol";
+import "./interfaces/ISecurityToken.sol";
 import "./interfaces/ISecurityTokenRegistry.sol";
+import "./modules/PermissionManager/IPermissionManager.sol";
 import "./storage/EternalStorage.sol";
 import "./libraries/Util.sol";
 import "./libraries/Encoder.sol";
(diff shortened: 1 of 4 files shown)


contracts/SecurityTokenRegistry.sol
```diff
@@ -453,42 +460,107 @@ contract SecurityTokenRegistry is ISecurityTokenRegistry, EternalStorage {
...
     }
 
+    function _ownerInToken(bytes32 _ticker, bool _allTokens, address _owner) internal view returns(address) {
+        address token = getAddressValue(Encoder.getKey("tickerToSecurityToken", Util.bytes32ToString(_ticker)));
+        if (token != address(0)) {
+            if (_allTokens || IOwnable(token).owner() == _owner) {
+                return token;
+            }
+        }
+        return token;
```
maxsam4: Shouldn't we return `address(0)` here?

maxsam4 (changes requested):

contracts/SecurityTokenRegistry.sol
satyamakgec: It will return automatically the `address(0)` I think as you read data from the eternal storage.

satyamakgec (approved):

contracts/SecurityTokenRegistry.sol
maxsam4: I meant for the case when token exists but the owner is someone else.

maxsam4 (approved):

PR #553, opened 2019-02-14
Upgradable modules

adamdossa:
Clean up interface vs. implementation using public getters in module factories.
Add UpgradableModuleFactory and use it consistently across upgradable modules.
Allow in-place upgrading of modules by token issuers.
Disallow changing of module types
Add fixed usage fees in POLY
Remove module instructions
Treat versions consistently in proxy based modules
Refactor module deployment to use base class code
Some refactoring of modifiers to internal functions
Clean up duplicated events in Security Token
Make LockUpTM upgradable (proxy deployment)
Allow all proxy based modules to be upgradable


contracts/mocks/MockCountTransferManager.sol
```diff
@@ -0,0 +1,95 @@
+pragma solidity ^0.5.0;
+
+import "../modules/TransferManager/TransferManager.sol";
+import "../modules/TransferManager/CountTransferManagerStorage.sol";
+
+/**
+ * @title Transfer Manager for limiting maximum number of token holders
+ */
+contract MockCountTransferManager is CountTransferManagerStorage, TransferManager {
```
satyamakgec: Couldn't we directly inherit the CountTransferManager directly?

contracts/modules/TransferManager/CountTransferManagerFactory.sol
```diff
@@ -43,41 +39,9 @@ contract CountTransferManagerFactory is ModuleFactory {
...
-        return "Allows an issuer to restrict the total number of non-zero token holders";
+        address countTransferManager = address(new CountTransferManagerProxy(logicContracts[latestVersion].version, msg.sender, IPolymathRegistry(polymathRegistry).getAddress("PolyToken"), logicContracts[latestVersion].logicContract));
+        _initializeModule(countTransferManager, _data);
+        return countTransferManager;
     }
 
-    /**
-     * @notice Get the tags related to the module factory
-     */
-    function getTags() external view returns(bytes32[] memory) {
```
satyamakgec, line 77: In every other module we are using the `tags()` function and `types()` function but here not. and we are not storing this info into the state variables of the modules but here we are storing them. 

we have the function `changeTags()` in `ModuleFactory.sol` which changes the tags of the factory but we call the `tags()` function it always returns the same value as it is not returning from the `tagsData` array (overriding the `ModuleFactory` functions). The same problem with `types()`

contracts/tokens/SecurityToken.sol
```diff
@@ -102,14 +102,6 @@ contract SecurityToken is ERC20, ERC20Detailed, ReentrancyGuard, RegistryUpdater
     event UpdateTokenDetails(string _oldDetails, string _newDetails);
     // Emit when the granularity get changed
     event GranularityChanged(uint256 _oldGranularity, uint256 _newGranularity);
-    // Emit when Module get archived from the securityToken
```
satyamakgec: This will remove the existence of these events from the SecurityToken ABI. So these events may not reflect in the etherscan until unless etherscan handling the events emitted by the liniking library. I am not sure but need to test this behavior.

satyamakgec (dismissed):
I think the `tm.sol` is accidentally added so need to remove it

contracts/libraries/TokenLib.sol
```diff
@@ -45,28 +48,38 @@ library TokenLib {
...
+        emit ModuleUnarchived(_moduleData.moduleTypes, _moduleData.module);
         _moduleData.isArchived = false;
     }
 
+    /**
+    * @notice Upgrades a module attached to the SecurityToken
+    * @param _moduleData Storage data
+    */
+    function upgradeModule(ModuleData storage _moduleData) public {
+        require(!_moduleData.isArchived, "Module archived");
```
maxsam4: I think we should allow upgrading Archived modules. This might be useful if a vulnerability is found. The issuer will be able to archive the module while we deploy a fix. I don't see any harm in upgrading archived modules.

contracts/modules/Experimental/Wallet/IWallet.sol
```diff
@@ -9,11 +9,13 @@ import "../../Module.sol";
  */
 contract IWallet is Module, Pausable {
```
maxsam4: Maybe let's rename this file to remove `I` prefix? I would like to reserve `I` for interface.

contracts/modules/UpgradableModuleFactory.sol
```diff
@@ -0,0 +1,88 @@
...
+ * @title Factory for deploying upgradable modules
+ */
+contract UpgradableModuleFactory is ModuleFactory {
+
+    event LogicContractSet(string _version, address _logicContract, bytes _logicData);
+
+    struct LogicContract {
+        string version;
+        address logicContract;
+        bytes logicData;
```
maxsam4, line 16: Maybe we should ask issuers to input `logicData`? If we store it like this, we won't be able to pass custom parameters.

contracts/modules/UpgradableModuleFactory.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // Mapping from version to logic contract
+    mapping (uint256 => LogicContract) logicContracts;
+
+    // Mapping from Security Token address to module version
+    mapping (address => uint256) moduleVersions;
+
+    // Mapping from Security Token address to module address
+    mapping (address => address) modules;
```
maxsam4: This will mess up when the issuer adds multiple of same module (multiple sto etc)

contracts/modules/UpgradableModuleFactory.sol
```diff
@@ -0,0 +1,88 @@
...
+        require(_logicContract != address(0), "Invalid address");
+        logicContracts[latestVersion].logicContract = _logicContract;
+        logicContracts[latestVersion].version = _version;
+    }
+
+    function setLogicContract(string calldata _version, address _logicContract, bytes calldata _logicData) external onlyOwner {
+        require(keccak256(abi.encodePacked(_version)) != keccak256(abi.encodePacked(logicContracts[latestVersion].version)), "Same version");
+        require(_logicContract != logicContracts[latestVersion].logicContract, "Same version");
+        require(_logicContract != address(0), "Invalid address");
+        latestVersion++;
```
maxsam4, line 68: We are now essentially splitting version into this integer version and the string human readable version. I think we should drop the string version from modules altogether. I don't see it serving any real purpose.

contracts/mocks/MockCountTransferManager.sol
adamdossa: Yeah - would be neater - will amend.

contracts/modules/TransferManager/CountTransferManagerFactory.sol
adamdossa: This should all be correct in the PR - oddly you seem to see an old version of the code!

Could you check again this point - it should be fixed in the commit from last night.

contracts/tokens/SecurityToken.sol
adamdossa: OK - thanks - will add a note to test this.

contracts/libraries/TokenLib.sol
adamdossa: Fair point - fixed.

contracts/modules/Experimental/Wallet/IWallet.sol
adamdossa: Done

contracts/modules/UpgradableModuleFactory.sol
adamdossa: My thinking here was:
  - logicData is for the upgrade process (determined by the factory owner) - e.g. if some data needs to be fixed as a part of the upgrade.
  - the issuer (i.e. ST) wouldn't be able to re-initialise the module as part of the upgrade process (which may break some internal logic on the module)

contracts/modules/UpgradableModuleFactory.sol
adamdossa: Good point... Will fix this.

contracts/modules/UpgradableModuleFactory.sol
adamdossa: I'm treating the integer string (`latestVersion`) as more of an internal variable to index logic contract deployments. Will have a think about this in a later PR though.

contracts/modules/UpgradableModuleFactory.sol
adamdossa: @maxsam4 Fixed - please review.

contracts/modules/UpgradableModuleFactory.sol
```diff
@@ -0,0 +1,109 @@
...
+    /* mapping (address => uint256) moduleVersions; */
+
+    // Mapping from Security Token address to module address
+    /* mapping (address => address) modules; */
+
+    // Mapping from Security Token address, to deployed proxy module address, to module version
+    mapping (address => mapping (address => uint256)) modules;
+
+    // Mapping of which security token owns a given module
+    mapping (address => address) moduleToSecurityToken;
```
maxsam4, line 32: Can't we just call the module to fetch st address?

adamdossa: Yep - probably could. Because of the risk around calling `upgradeToAndCall` on a module I wanted the factory to be very sure that the `upgrade` function was being called by the module owner (i.e. the address that initially deployed the module) without worrying about what a malicious module could potentially achieve by spoofing its owner. I can't think of a way this could be exploited though, assuming that the module being upgraded was not itself malicious, but thought it was safer this way.

maxsam4 (approved):

satyamakgec (approved):

PR #555, opened 2019-02-14
Add treasury wallet in global storage

satyamakgec:


contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -266,8 +266,10 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
                 tiers[i].mintedTotal = tiers[i].tokenTotal;
             }
         }
-        require(ISecurityToken(securityToken).mint(reserveWallet, tempReturned), "Minting Failed");
-        emit ReserveTokenMint(msg.sender, reserveWallet, tempReturned, currentTier);
+        address _wallet = (treasuryWallet == address(0) ? IDataStore(getDataStore()).getAddress(TREASURY) : treasuryWallet);
+        require(_wallet != address(0));
```
F-OBrien: Is there a case where an issuer would not want to automatically mint unsold tokens to a treasury wallet? In a STO implementation I am working on I have it so if the reserve wallet is set to address zero the STO will finalize but won't mint unsold tokens.
I've been trying to keep my STO up to date with changes in the dev 3.0.0 branch but this one would mean I'd require both the STO configured treasury wallet and global treasury wallet to be zero for minting unsold tokens to be disabled. Depending on what else the global treasury wallet is used for this might make disabling minting unsold tokens in this manner impossible. 
This only impacts my STO contract currently so I hope you don't mind my comment here. I guess the issuer always has the option to burn minted tokens they don't want but though I'd mention this point for you to consider. I may just add a function for finalizeMintUnsoldToTreasury and finalizeWitoutMinting instead to maintain this option while aligning with this change (once merged to Dev-3.0.0).
Edit: Actually I might just add a bool input to my finalize function to mintUnsold or not instead.

satyamakgec: @F-OBrien The Idea is very same as the earlier implementation has, Issuer has to mint the unsold token to reserve/local treasury wallet/global treasury wallet. Technically yes you could work with one more boolean variable. If we think of a more legal perspective as issuer committed to issuing `X` amount of supply in STO (on legal documents) then issuer has to mint that much amount of tokens by considering that issuer can burn those in another transaction by giving some legal statement. It will help to maintain the ledger of legal documents. I am not quite sure though but it looks cleaner.

contracts/modules/ModuleStorage.sol
```diff
@@ -13,6 +13,8 @@ contract ModuleStorage {
 
     bytes32 public constant FEE_ADMIN = "FEE_ADMIN";
 
+    bytes32 public constant TREASURY = "TREASURY_WALLET";
```
maxsam4: Please store the hash of "TREASURY_WALLET" and not "TREASURY_WALLET". All keys should be keccak hashes. Also, maybe declare it has internal? We don't need a getter for that in all the modules. It will be a waste of code size.

maxsam4 (changes requested):

contracts/SecurityTokenRegistry.sol
```diff
@@ -505,23 +507,27 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
         require(_tickerOwner(ticker) == msg.sender, "Not authorised");
         /*solium-disable-next-line security/no-block-members*/
         require(getUintValue(Encoder.getKey("registeredTickers_expiryDate", ticker)) >= now, "Ticker gets expired");
-        (uint256 _usdFee, uint256 _polyFee) = _takeFee(STLAUNCHFEE);
+        _deployToken(_name, ticker, _tokenDetails, msg.sender, _divisible, _treasuryWallet); 
+    }
 
+    function _deployToken(string memory name, string memory ticker, string memory tokenDetails, address issuer, bool divisible, address wallet) internal {
```
adamdossa: Minor but generally parameters should start with _

contracts/datastore/DataStore.sol
```diff
@@ -16,13 +16,17 @@ contract DataStore is DataStoreStorage, IDataStore {
     event SecurityTokenChanged(address indexed _oldSecurityToken, address indexed _newSecurityToken);
 
     modifier onlyAuthorized() {
-        bool isOwner = msg.sender == IOwnable(address(securityToken)).owner();
-        require(isOwner ||
-            securityToken.isModule(msg.sender, DATA_KEY) ||
-            securityToken.checkPermission(msg.sender, address(this), MANAGEDATA),
+        _isAuthorized();
```
adamdossa: Any reason not to just replace the whole modifier with a call to _isAuthorized()? Not convinced that modifiers are easier to read than a function call...

contracts/modules/Checkpoint/DividendCheckpoint.sol
```diff
@@ -35,33 +35,19 @@ contract DividendCheckpoint is DividendCheckpointStorage, ICheckpoint, Module {
...
     /**
      * @notice Function used to change wallet address
      * @param _wallet Ethereum account address to receive reclaimed dividends and tax
      */
     function changeWallet(address payable _wallet) external onlyOwner {
-        _setWallet(_wallet);
-    }
-
-    function _setWallet(address payable _wallet) internal {
         require(_wallet != address(0));
```
adamdossa: Wondering whether we should allow setting of address(0) now which would go back to using the dataStore default?

contracts/modules/Checkpoint/ERC20DividendCheckpoint.sol
```diff
@@ -272,8 +274,10 @@ contract ERC20DividendCheckpoint is ERC20DividendCheckpointStorage, DividendChec
         Dividend storage dividend = dividends[_dividendIndex];
         uint256 remainingWithheld = dividend.totalWithheld.sub(dividend.totalWithheldWithdrawn);
         dividend.totalWithheldWithdrawn = dividend.totalWithheld;
-        require(IERC20(dividendTokens[_dividendIndex]).transfer(wallet, remainingWithheld), "transfer failed");
-        emit ERC20DividendWithholdingWithdrawn(wallet, _dividendIndex, dividendTokens[_dividendIndex], remainingWithheld);
+        address _wallet = getTreasuryWallet();
```
adamdossa: I'd prefer using _ consistently for function parameters

contracts/modules/Checkpoint/EtherDividendCheckpoint.sol
```diff
@@ -198,8 +198,10 @@ contract EtherDividendCheckpoint is DividendCheckpoint {
         Dividend storage dividend = dividends[_dividendIndex];
         dividend.reclaimed = true;
         uint256 remainingAmount = dividend.amount.sub(dividend.claimedAmount);
-        wallet.transfer(remainingAmount);
-        emit EtherDividendReclaimed(wallet, _dividendIndex, remainingAmount);
+        address payable _wallet = getTreasuryWallet();
```
adamdossa: Would prefer _ to be used consistently for input params

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -104,7 +104,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
         uint256 _minimumInvestmentUSD,
         FundRaiseType[] memory _fundRaiseTypes,
         address payable _wallet,
-        address _reserveWallet,
+        address _treasuryWallet,
```
adamdossa: Should we just remove this from configure similar to Dividend modules? Then allow them to override it in a subsequent call to changeAddresses if they want to?

contracts/tokens/STFactory.sol
```diff
@@ -42,6 +43,7 @@ contract STFactory is ISTFactory {
         );
         //NB When dataStore is generated, the security token address is automatically set via the constructor in DataStoreProxy.
         newSecurityToken.changeDataStore(dataStoreFactory.generateDataStore(address(newSecurityToken)));
+        newSecurityToken.changeTreasuryWallet(_treasuryWallet);
```
adamdossa, line 49: Bit cheaper to only do this if _treasuryWallet != 0

contracts/datastore/DataStore.sol
satyamakgec: The only reason to keep as it is to keep functions readable.

contracts/modules/STO/USDTieredSTO.sol
satyamakgec: I keep it because of backward compatibility, but anyway I think we have to deploy everything for 3.0 release so I will make that change.

contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -137,8 +135,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
      * @param _minimumInvestmentUSD overall minimum investment limit
      */
     function modifyLimits(uint256 _nonAccreditedLimitUSD, uint256 _minimumInvestmentUSD) external onlyOwner {
-        /*solium-disable-next-line security/no-block-members*/
-        require(now < startTime, "STO already started");
+        _isSTOStarted();
```
maxsam4: As the check is only one line, It might be better to define it as a function modifier.

satyamakgec: To save the size, I had to make function instead of modifier. it saves .20 KB

maxsam4 (approved):

contracts/interfaces/ISecurityTokenRegistry.sol
```diff
@@ -13,6 +13,22 @@ interface ISecurityTokenRegistry {
...
 
+    /**
+     * @notice Deploys an instance of a new Security Token and records it to the registry
+     * @param _name is the name of the token
+     * @param _ticker is the ticker symbol of the security token
+     * @param _tokenDetails is the off-chain details of the token
+     * @param _divisible is whether or not the token is divisible
+     * @param _treasuryWallet Ethereum address which will holds the STs.
+     */
+    function generateSecurityTokenWithTreasury(
```
maxsam4: Name need to be updated.

maxsam4 (changes requested):
Can you please also update the changelog.
Rest of the PR looks good to me.

maxsam4 (approved):

adamdossa:
@satyamakgec As per discussion this morning, please can you add treasury wallet to module config files (where it existed previously) but with the logic that address(0) is valid (and corresponds to using the DS version).

contracts/modules/Checkpoint/ERC20/ERC20DividendCheckpoint.sol
```diff
@@ -257,7 +257,7 @@ contract ERC20DividendCheckpoint is ERC20DividendCheckpointStorage, DividendChec
         dividends[_dividendIndex].reclaimed = true;
         Dividend storage dividend = dividends[_dividendIndex];
         uint256 remainingAmount = dividend.amount.sub(dividend.claimedAmount);
-        require(IERC20(dividendTokens[_dividendIndex]).transfer(wallet, remainingAmount), "transfer failed");
+        require(IERC20(dividendTokens[_dividendIndex]).transfer(getTreasuryWallet(), remainingAmount), "transfer failed");
```
adamdossa, line 260: For good measure do you think we should be checking that `getTreasuryWallet()` doesn't return address(0)? I agree that with the current code it can't, but in some future model if there was a bug, perhaps the datastore treasury wallet could be over-written with 0, and we wouldn't want to send funds there under any circumstances.

contracts/modules/Experimental/Wallet/VestingEscrowWallet.sol
```diff
@@ -107,10 +109,20 @@ contract VestingEscrowWallet is VestingEscrowWalletStorage, Wallet {
...
     }
 
+    /**
+     * @notice Returns the treasury wallet address
+     */
+    function getTreasuryWallet() public view returns(address) {
+        if (treasuryWallet == address(0)) 
+            return IDataStore(getDataStore()).getAddress(TREASURY);
+        else
+            return treasuryWallet;
```
adamdossa, line 125: Similar to dividends, could check that final returned value is not 0.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -41,7 +41,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
         uint256 _rate
     );
     event ReserveTokenMint(address indexed _owner, address indexed _wallet, uint256 _tokens, uint256 _latestTier);
-    event SetAddresses(address indexed _wallet, address indexed _reserveWallet, address[] _usdTokens);
+    event SetAddresses(address indexed _wallet, address indexed treasuryWallet, address[] _usdTokens);
```
adamdossa: _treasuryWallet for consistency

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -50,6 +50,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
         uint256[] _tokensPerTierTotal,
         uint256[] _tokensPerTierDiscountPoly
     );
+    event SetTreasuryWallet(address _oldWallet, address _newWallet);
```
adamdossa, line 53: Is this event ever used?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -245,16 +246,16 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
             require(_usdTokens[i] != address(0), "Invalid USD token");
             usdTokenEnabled[_usdTokens[i]] = true;
         }
-        emit SetAddresses(wallet, reserveWallet, _usdTokens);
-    }
+        emit SetAddresses(wallet, treasuryWallet, _usdTokens);
```
adamdossa: Perhaps should emit SetTreasuryWallet with the new value for the treasury wallet (i.e. default DS value or treasuryWallet if non-zero)?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -271,8 +272,9 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
                 tiers[i].mintedTotal = tiers[i].tokenTotal;
             }
         }
-        ISecurityToken(securityToken).issue(reserveWallet, tempReturned, "");
-        emit ReserveTokenMint(msg.sender, reserveWallet, tempReturned, currentTier);
+        address walletAddress = (treasuryWallet == address(0) ? IDataStore(getDataStore()).getAddress(TREASURY) : treasuryWallet);
+        ISecurityToken(securityToken).issue(walletAddress, tempReturned, "");
```
adamdossa, line 278: Should check that walletAddress != 0

contracts/tokens/STFactory.sol
```diff
@@ -45,6 +46,7 @@ contract STFactory is ISTFactory {
...
```
adamdossa, line 49: What's the reason to not add this to the token constructor directly?

adamdossa (changes requested):

contracts/modules/Checkpoint/ERC20/ERC20DividendCheckpoint.sol
satyamakgec: on this measure, global treasury wallet is only be changed using the `changeTreasuryWallet()` function that is present in the SecurityToken and we have the 0 address check init.  So it doesn't require any further check.  And yes there is no harm to add one more check.

contracts/tokens/STFactory.sol
satyamakgec: because datastore doesn't aware of the security token address at that time.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
satyamakgec: I don't know how it is showing on my local it is correct.

adamdossa (approved):

PR #557, opened 2019-02-19
Dependency cleanup

maxsam4:
Dependency cleanup


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "polymath-core",
-  "version": "2.0.0",
+  "version": "3.0.0",
```
satyamakgec, line 3: I prefer the bump of the version during the release.

maxsam4: We have already bumped 2.1 version. It helps the dApp guys. I am fine with any though.

satyamakgec: Ok then leave as it is

satyamakgec (approved):

PR #559, opened 2019-02-20
GTM KYC during transfer

maxsam4:
Allows authorized address to KYC/whitelist multiple addresses during a transfer.


maxsam4:
@adamdossa @satyamakgec The logic and already added test cases are ready for review. I will add a couple more tests cases before merging to assert that it fails when it should fail.

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
```diff
@@ -66,9 +67,9 @@ contract SignedTransferManager is TransferManager {
     */
     function executeTransfer(address _from, address _to, uint256 _amount, bytes calldata _data) external onlySecurityToken returns(Result) {
         (Result success, ) = verifyTransfer(_from, _to, _amount, _data);
-        if (success == Result.VALID) {
+        if (success == Result.VALID && _data.length > 32) {
```
satyamakgec, line 70: do we really need the data length check here?

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
```diff
@@ -43,10 +43,11 @@ contract SignedTransferManager is TransferManager {
     function checkSignatureValidity(bytes calldata _data) external view returns(bool) {
         address targetAddress;
         uint256 nonce;
+        uint256 validFrom;
         uint256 expiry;
         bytes memory signature;
-        (targetAddress, nonce, expiry, signature) = abi.decode(_data, (address, uint256, uint256, bytes));
-        if (targetAddress != address(this) || expiry < now || signature.length == 0 || _checkSignatureIsInvalid(signature))
+        (targetAddress, nonce, validFrom, expiry, signature) = abi.decode(_data, (address, uint256, uint256, uint256, bytes));
```
satyamakgec, line 49: what is the idea behind using `validFrom`?

satyamakgec (approved):

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
maxsam4: `validFrom` is the date from which the signature is valid. This is the standard data structure we decided to use for all modules that take data while transfer.

contracts/modules/Experimental/TransferManager/SignedTransferManager.sol
maxsam4: Yes, if we try to decode empty data, it will revert.

satyamakgec: but this will automatically check in verifyTransfer

maxsam4: but it is decoded here also to invalidate the signature.

adamdossa (approved):

PR #560, opened 2019-02-20
Remove the investor array and add isVolRestricted flag for investors

satyamakgec:


maxsam4 (commented):
Please update the Flags document. Rest LGTM.

PR #564, opened 2019-02-20
Merge 2.1 / 2.2 functionality

adamdossa:


satyamakgec:
@adamdossa I reduced the size for now but at the end, we are gonna use the proxy approach which will reduce the size of the factory automatically

maxsam4 (approved):

PR #567, opened 2019-02-21
Add version to `getSecurityTokenData` (referenced directly from token using getVersion)

satyamakgec:


maxsam4 (approved):

adamdossa (approved):

maxsam4:
This is now a combination of all the pending PRs.
Combined the PRs so that we don't have to wait an eternity for the CI to finish one by one :)

PR #568, opened 2019-02-21
Allows explicit token factory version in generateSecurityToken()

satyamakgec:


contracts/SecurityTokenRegistry.sol
```diff
@@ -487,41 +487,53 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
         string calldata _ticker,
         string calldata _tokenDetails,
-        bool _divisible
+        bool _divisible,
+        uint256 _protocolVersion
     )
         external
         whenNotPausedOrOwner
     {
+        uint256 protocolVersion = _protocolVersion;
```
maxsam4, line 505: Why do we need `protocolVersion`? I think we can just manipulate `_protocolVersion`.

contracts/SecurityTokenRegistry.sol
```diff
@@ -487,41 +487,53 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
         set(statusKey, true);
         require(_tickerOwner(ticker) == msg.sender, "Not authorised");
         /*solium-disable-next-line security/no-block-members*/
         require(getUintValue(Encoder.getKey("registeredTickers_expiryDate", ticker)) >= now, "Ticker expired");
-        (uint256 _usdFee, uint256 _polyFee) = _takeFee(STLAUNCHFEE);
+        _deployToken(_name, ticker, _tokenDetails, msg.sender, _divisible, protocolVersion); 
+    }
 
-        address newSecurityTokenAddress = ISTFactory(getAddressValue(Encoder.getKey("protocolVersionST", getUintValue(Encoder.getKey("latestVersion"))))).deployToken(
+    function _deployToken(string memory _name, string memory _ticker, string memory _tokenDetails, address _issuer, bool _divisible, uint256 _protocolVersion) internal {
```
maxsam4: Please break this definition into multiple lines.

contracts/SecurityTokenRegistry.sol
```diff
@@ -487,41 +487,53 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
         );
 
         /*solium-disable-next-line security/no-block-members*/
-        _storeSecurityTokenData(newSecurityTokenAddress, ticker, _tokenDetails, now);
-        set(Encoder.getKey("tickerToSecurityToken", ticker), newSecurityTokenAddress);
+        _storeSecurityTokenData(newSecurityTokenAddress, _ticker, _tokenDetails, now);
+        set(Encoder.getKey("tickerToSecurityToken", _ticker), newSecurityTokenAddress);
         /*solium-disable-next-line security/no-block-members*/
-        emit NewSecurityToken(ticker, _name, newSecurityTokenAddress, msg.sender, now, msg.sender, false, _usdFee, _polyFee);
+        emit NewSecurityToken(_ticker, _name, newSecurityTokenAddress, msg.sender, now, msg.sender, false, _usdFee, _polyFee);
```
maxsam4: I think `protocolVersion` should also be part of the event.

maxsam4 (commented):
Please update the STR interface as well.

contracts/SecurityTokenRegistry.sol
satyamakgec: I did the same but I was getting lvalue error so I use that approach.

maxsam4 (approved):

contracts/SecurityTokenRegistry.sol
adamdossa: Yep - calldata is readonly and this is an external function.

contracts/SecurityTokenRegistry.sol
adamdossa: Agreed.

contracts/STRGetter.sol
```diff
@@ -222,7 +222,7 @@ contract STRGetter is EternalStorage {
     /**
      * @notice Gets Protocol version
      */
-    function getProtocolVersion() public view returns(uint8[] memory) {
+    function getCurrentProtocolVersion() public view returns(uint8[] memory) {
```
adamdossa: maybe LatestProtocolVersion to be consistent.

contracts/SecurityTokenRegistry.sol
```diff
@@ -532,18 +554,20 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
      */
     function modifySecurityToken(
         string calldata _name,
         string calldata _ticker,
         address _owner,
         address _securityToken,
         string calldata _tokenDetails,
-        uint256 _deployedAt
+        uint256 _deployedAt,
+        uint256 _protocolVersion
```
adamdossa: Do you think this is needed for the modifyST? It may be sufficient to just set this to 0 in the event. This data would only be as good as the value passed in, and in practice we'll care more about what the deployed version of the token (from the token itself) is, rather than parsing these events.

adamdossa:
@satyamakgec left a couple of minor comments.

contracts/SecurityTokenRegistry.sol
satyamakgec: Yes, I was also thinking the same. But I did it because of consistency. I will make the change anyway.

PR #572, opened 2019-02-26
GTM logic refactored

maxsam4:
GTM logic has been simplified and made more flexible.
Some work will be required by the dApp team to incorporate the changes.


contracts/proxy/GeneralTransferManagerProxy.sol
```diff
@@ -26,6 +26,7 @@ contract GeneralTransferManagerProxy is GeneralTransferManagerStorage, ModuleSto
     {
         require(_implementation != address(0), "Implementation address should not be 0x");
         _upgradeTo(_version, _implementation);
+        _setDefaults();
```
adamdossa: Wonder whether this should be in the GTM initialisation function instead? Possibly in a future upgraded GTM this function may change.

adamdossa: I guess it only gets called the first time the proxy is created and the logic contract is set (not on future upgradeToAndCall calls) so this is fine.

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -150,67 +107,133 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+    /**
+    * @notice Modifies the successful checks required for transfers.
+    * @param _transferType Type of transfer (0 = General, 1 = Issuance, 2 = Redemption)
+    * @param _fromValidKYC Defines if KYC is required for the sender
+    * @param _toValidKYC Defines if KYC is required for the receiver
+    * @param _fromRestricted Defines if transfer time restriction is checked for the sender
+    * @param _toRestricted Defines if transfer time restriction is checked for the receiver
+    */
+    function modifyTransferRequirementsMulti(
+        uint256[] memory _transferType,
```
satyamakgec: a plural form should be used here

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -150,67 +107,133 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+    }
+
+    function _modifyTransferRequirements(
+        uint256 _transferType,
+        bool _fromValidKYC,
+        bool _toValidKYC,
+        bool _fromRestricted,
+        bool _toRestricted
+    ) internal {
+        transferRequirements[_transferType] =
```
satyamakgec, line 221: can add a sanity check on `transferType`

satyamakgec (commented):
The approach is good on the contract side but it will little bit harden the dApp work. dApp team needs to create some scenarios (combination of true or false) when issuer wants to change in the transfers flag.
And redemption/ allowAllWhitelistTransfers will little bit costlier.

maxsam4:
@satyamakgec `allowAllWhitelistTransfers` will actually be cheaper. Ones less SLOAD is required in the new logic. 

Only `allowAllTransfers` and `allowAllBurnTransfers` will be slightly more expensive but I don't think anyone will use them in practice.

satyamakgec (approved):

adamdossa (approved):

PR #578, opened 2019-03-04
Add weighted voting module

satyamakgec:


contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+ * @notice In this module every token holder has voting right (Should be greater than zero)
+ * Tally will be calculated as per the weight (balance of the token holder)
+ */
+contract WeightedVoteCheckpoint is ICheckpoint, Module {
+    using SafeMath for uint256;
+
+    struct Ballot {
+        uint256 checkpointId;
+        uint256 totalSupply;
+        uint256 startTime;
```
maxsam4: Most of these variables can safely be stored in uint64.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+        uint256 totalNumVotes;
+        uint256 totalProposals;
+        bool isActive;
+        mapping(uint256 => uint256) proposalToVote;
+        mapping(address => Vote) voteByAddress;
+    }
+
+    Ballot[] ballots;
+
+    struct Vote {
```
maxsam4: Do we really need this at all? I think all this information can be encapsulated in the Ballot itself.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+    function getResults(uint256 _ballotId) external view returns (
+        uint256[] memory,
+        uint256 remainingTime,
+        uint256 totalVotes
+    ) {
+        if (_ballotId < ballots.length) {
+            Ballot storage ballot = ballots[_ballotId];
+            uint256[] memory abstain = new uint256[](ballot.totalProposals);
+            for (uint256 i = 0; i < ballot.totalProposals; i++) {
+                abstain[i] = ballot.proposalToVote[i+1];
```
maxsam4: Abstain? This is the net proposal vote weight, right?

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+     * @param _ballotId The index of the target ballot
+     * @param _proposalId Id of the proposal which investor want to vote for proposal
+     */
+    function castVote(uint256 _ballotId, uint256 _proposalId) external {
+        require(_ballotId < ballots.length, "Incorrect ballot Id");
+        Ballot storage ballot = ballots[_ballotId];
+
+        uint256 weight = ISecurityToken(securityToken).balanceOfAt(msg.sender, ballot.checkpointId);
+        require(weight > 0, "weight should be > 0");
+        require(ballot.totalProposals >= _proposalId && _proposalId > 0, "Incorrect proposals Id");
```
maxsam4, line 106: Maybe we should allow voting for proposal Id 0 and consider that as reject vote. i.e the voter is against all the proposals.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+            ballot.checkpointId,
+            ballot.totalSupply,
+            ballot.startTime,
+            ballot.endTime,
+            ballot.totalNumVotes,
+            ballot.totalProposals,
+            ballot.isActive
+        );
+    }
+
```
maxsam4, line 206: I think we should add a getter to return the proposal selected by the voter. 
It should take ballot id and voter address as input and return the proposal selected.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+        return bytes4(0);
+    }
+
+    /**
+     * @notice Allows the token issuer to create a ballot
+     * @param _duration The duration of the voting period in seconds
+     * @param _noOfProposals Number of proposals
+     */
+    function createBallot(uint256 _duration, uint256 _noOfProposals) external {
+        _onlySecurityTokenOwner();
```
adamdossa: probably should use a `withPerm` of operator

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @notice Allows the token issuer to create a ballot with custom settings
+     * @param _startTime Start time of the voting period in Unix Epoch time
+     * @param _endTime End time of the voting period in Unix Epoch time
+     * @param _checkpointId Index of the checkpoint to use for token balances
+     * @param _noOfProposals Number of proposals
+     */
+    function createCustomBallot(uint256 _startTime, uint256 _endTime, uint256 _checkpointId, uint256 _noOfProposals) external {
+        _onlySecurityTokenOwner();
+        _createCustomBallot(_startTime, _endTime, _checkpointId, _noOfProposals);
```
adamdossa, line 92: We could explicitly check that `_checkpointId` is valid for a better error message (although `totalSupplyAt` would fail with an invalid checkpoint).

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
adamdossa: I think we should allow voting for proposal 0 - not sure why we wouldn't index proposals from 0? I don't think proposal 0 should have any specific special meaning though.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
adamdossa: Agree `abstain` seems like a bad variable name - perhaps voteWeighting or something. Still confused why we are doing the off by 1 stuff with `ballot.proposalToVote[i+1]`

satyamakgec: Yep I agree to, abstain is the bad word, I am making off by 1 because proposals always will be greater than one. at the 0th place, there will be no data.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
satyamakgec: How do you think it can be encapsulated in ballot? It is the information specific to the voter.

maxsam4: It has Vote Time (not needed imo), Vote weight (not needed separately, we can just store the sum of vote weights) and vote. We can keep a mapping of address -> vote in the Ballot and eliminate the need of this struct.

satyamakgec: The Vote struct is used to keep track of the data with respect to the voter. So if we want to show the data `who votes whom`. this struct will be useful and the timestamp is used to understand the voting campaign impact according to the timing of votes.

maxsam4: As I mentioned earlier, we can just keep address -> vote mapping in the ballot to keep track of who voted for whom. I don't see the need for this struct. As far as timestamp goes, we can just use event/transaction timestamp for analysis. This analysis will be done offchain so we don't need to store timestamp like that onchain.

satyamakgec:
Everything is fixed only one thing remains i.e `proposalId` starts from 0. I keep that as it is because if I start with it 0 then there is no way to restrict the user to vote again until unless I will use another storage variable to keep track of it. IMO it is best over introducing another variable to do it.

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        uint256 weight = ISecurityToken(securityToken).balanceOfAt(msg.sender, ballot.checkpointId);
+        require(weight > 0, "weight should be > 0");
+        require(ballot.totalProposals >= _proposalId && _proposalId > 0, "Incorrect proposals Id");
+        require(now >= ballot.startTime && now <= ballot.endTime, "Voting period is not active");
+        require(ballot.voteByAddress[msg.sender] == 0, "Token holder has already voted");
+        require(ballot.isActive == uint8(1), "Ballot is not active");
+
+        ballot.voteByAddress[msg.sender] = _proposalId;
+        ballot.totalNumVotes = uint64(uint256(ballot.totalNumVotes).add(uint256(1)));
```
maxsam4: Minor optimization: we don't need to use safemath.add here because it can never overflow.
(Adding 1 to a 64 bit number can never result in overflow of a 256 bit number). We don't even need to check for overflow of 64bits because I doubt any ballot is gonna have 18,446,744,073,709,551,616 votes cast :).

test/zc_weighted_vote_checkpoint.js
```diff
@@ -0,0 +1,409 @@
...
+
+        STFactory:                         ${I_STFactory.address}
+        GeneralTransferManagerFactory:     ${I_GeneralTransferManagerFactory.address}
+
+        WeightedVoteCheckpointFactory:     ${I_WeightedVoteCheckpointFactory.address}
+        -----------------------------------------------------------------------------
+        `);
+    });
+
+    describe("\t\t Test case for Weighted Vote Checkpoint module \n", async() => {
```
maxsam4, line 126: Why are there new line and tab characters everywhere?

satyamakgec: It is only for better presentation

maxsam4 (approved):

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+        uint56 totalProposals;
+        uint8 isActive;
+        mapping(uint256 => uint256) proposalToVote;
+        mapping(address => uint256) voteByAddress;
+    }
+
+    Ballot[] ballots;
+
+    event BallotCreated(uint256 _startTime, uint256 _endTime, uint256 _ballotId, uint256 _checkpointId, uint256 _noOfProposals);
+    event VoteCasted(uint256 indexed _ballotId, uint256 indexed _proposalId, address indexed _investor, uint256 _weight);
```
adamdossa: Minor - should be VoteCast, not VoteCasted

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+    function createBallot(uint256 _duration, uint256 _noOfProposals) external withPerm(ADMIN) {
+        require(_duration > 0, "Incorrect ballot duration.");
+        uint256 checkpointId = ISecurityToken(securityToken).createCheckpoint();
+        uint256 endTime = now.add(_duration);
+        _createCustomBallot(now, endTime, checkpointId, _noOfProposals);
+    }
+
+    function _createCustomBallot(uint256 _startTime, uint256 _endTime, uint256 _checkpointId, uint256 _noOfProposals) internal {
+        require(_noOfProposals > 1, "Incorrect proposals no");
+        require(_endTime > _startTime, "Times are not valid");
```
adamdossa, line 66: Should we require(_startTime >= now) as well?

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+        Ballot storage ballot = ballots[_ballotId];
+
+        uint256 weight = ISecurityToken(securityToken).balanceOfAt(msg.sender, ballot.checkpointId);
+        require(weight > 0, "weight should be > 0");
+        require(ballot.totalProposals >= _proposalId && _proposalId > 0, "Incorrect proposals Id");
+        require(now >= ballot.startTime && now <= ballot.endTime, "Voting period is not active");
+        require(ballot.voteByAddress[msg.sender] == 0, "Token holder has already voted");
+        require(ballot.isActive == uint8(1), "Ballot is not active");
+
+        ballot.voteByAddress[msg.sender] = _proposalId;
```
adamdossa: Minor - should probably be proposalByAddress perhaps? i.e. it is not a count of votes by address (which we don't store as it can be reconstructed from balanceOfAt, but a mapping of which proposal was chosen).

adamdossa: or maybe investorToProposal if you want to be consistent with proposalToVote

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /**
+     * @notice Allows the token issuer to set the active stats of a ballot
+     * @param _ballotId The index of the target ballot
+     * @param _isActive The bool value of the active stats of the ballot
+     * @return bool success
+     */
+    function changeBallotStatus(uint256 _ballotId, bool _isActive) external withPerm(ADMIN) {
+        require(uint64(now) < ballots[_ballotId].endTime, "Already ended");
```
adamdossa: Should be <= to be consistent with castVote

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+     * @return uint256 voteWeighting
+     * @return uint256 tieWith
+     * @return uint256 winningProposal
+     * @return uint256 remainingTime
+     * @return uint256 totalVotes
+     
+     */
+    function getBallotResults(uint256 _ballotId) external view returns (
+        uint256[] memory,
+        uint256[] memory,
```
adamdossa: Any reason we name some of the return params, but not others?

contracts/modules/Experimental/Checkpoints/WeightedVoteCheckpoint.sol
```diff
@@ -0,0 +1,222 @@
...
+        uint256 totalSupply;
+        uint64 startTime;
+        uint64 endTime;
+        uint64 totalNumVotes;
+        uint56 totalProposals;
+        uint8 isActive;
+        mapping(uint256 => uint256) proposalToVote;
+        mapping(address => uint256) voteByAddress;
+    }
+
```
adamdossa, line 26: This is really picky (sorry), but I think these variable names could be cleaned up a bit to make things clearer. I'd suggest:
```
    struct Ballot {
        uint256 checkpointId;
        uint256 totalSupply;
        uint64 startTime;
        uint64 endTime;
        uint64 totalVoters;
        uint56 totalProposals;
        uint8 isActive;
        mapping(uint256 => uint256) proposalCount; //or proposalToVotes
        mapping(address => uint256) investorProposal; //or investorToProposal
    }
```

PR #583, opened 2019-03-06
STO fixes

maxsam4:
- Renamed ISTO to STO
- Added STO interface
- Removed redundant definition of wallet in USDTieredSTO
- Removed unused functions
- Made a few functions external
- Added a couple of sanity checks
- Refund remaining amount in Capped STO rather than reverting when the cap is reached by that purchase.

### DIFF
--- a/contracts/modules/STO/USDTieredSTO.sol
+++ b/contracts/modules/STO/USDTieredSTO.sol
@@ -1,6 +1,6 @@
 pragma solidity ^0.4.24;
 
-import "./ISTO.sol";
+import "./STO.sol";
 import "../../interfaces/ISecurityToken.sol";
 import "../../interfaces/IOracle.sol";
 import "../../RegistryUpdater.sol";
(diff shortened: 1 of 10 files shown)


contracts/modules/STO/USDTieredSTO.sol
```diff
@@ -262,7 +262,7 @@ contract USDTieredSTO is USDTieredSTOStorage, ISTO, ReentrancyGuard {
         }
         usdTokens = _usdTokens;
         for(i = 0; i < _usdTokens.length; i++) {
-            require(_usdTokens[i] != address(0), "Invalid USD token");
+            require(_usdTokens[i] != address(0) && _usdTokens[i] != address(0), "Invalid USD token");
```
satyamakgec: isn't `address(0)` need to replace with `polyToken`.
`require(_usdTokens[i] != address(0) && _usdTokens[i] != address(polyToken), "Invalid USD token");`

satyamakgec (changes requested):

contracts/modules/STO/USDTieredSTO.sol
maxsam4: Oh yes. It was a typo.

satyamakgec (approved):

PR #584, opened 2019-03-06
[docs] Force overrite existing solc

maxsam4:
Should fix the docs building process


satyamakgec (approved):
Hopefully, it is running fine :)

PR #589, opened 2019-03-07
PLCR voting module

satyamakgec:


contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+
+    struct Vote {
+        uint256 voteOption;
+        bytes32 secretVote;
+        bool isRevealed;
+    }
+
+    Ballot[] ballots;
+
+    event VoteCommited(address indexed _voter, uint256 _weight, uint256 _ballotId, bytes32 _secretVote);
```
maxsam4: I think marking `_ballotId` as `indexed` will be useful.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+        uint256 indexed _ballotId,
+        uint256 _commitDuration,
+        uint256 _revealDuration,
+        uint256 _noOfProposals,
+        uint256 _startTime,
+        uint256 _proposedQuorum,
+        uint256 _totalSupply,
+        uint256 _checkpointId
+    );
+    event VoteRevealed(address _voter, uint256 _weight, uint256 indexed _ballotId, uint256 _choiceOfProposal, uint256 _salt, bytes32 _secretVote);
```
maxsam4: I think marking `_voter` as `indexed` will be useful.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+    event BallotStatusChanged(uint256 _ballotId, bool _newStatus);
```
maxsam4: I think marking `_ballotId` as `indexed` will be useful.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+        uint256 _startTime
+    )
+        internal
+    {
+        // Sanity checks
+        _validValueCheck(_commitDuration);
+        _validValueCheck(_revealDuration);
+        _validValueCheck(_proposedQuorum);
+        require(_startTime >= now, "Invalid start time");
+        require(_totalProposals > 1, "totalProposals should be > 1");
```
maxsam4: For consistency, reason strings should mention what went wrong rather than what is the valid input.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+    struct Ballot {
+        uint256 quorum;       // Should be a multiple of 10 ** 16
+        uint256 totalSupply;
+        uint256 checkpointId;
+        uint64 commitDuration;
+        uint64 revealDuration;
+        uint64 startTime;
+        uint24 totalProposals;
+        uint32 totalVotes;
+        uint8 isActive;
```
maxsam4: Solidity can now pack different types of variables together so we can define `isActive` as bool and still have it packed. A bool is equivalent to uint8.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+        mapping(uint256 => uint256) weightedVote;
+        mapping(address => Vote) proposalToVote;
+    }
+
+    struct Vote {
+        uint256 voteOption;
+        bytes32 secretVote;
+        bool isRevealed;
```
maxsam4: We don't need `isRevealed`. Once the voter reveals their vote, we should delete `secretVote`. This will make both commit and reveal transactions cheaper.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,313 @@
...
+
+contract PLCRVotingCheckpoint is Module {
+
+    using SafeMath for uint256;
+
+    enum Stage { PREP, COMMIT, REVEAL, RESOLVED }
+
+    struct Ballot {
+        uint256 quorum;       // Should be a multiple of 10 ** 16
+        uint256 totalSupply;
```
maxsam4: We don't need to store `totalSupply`. We can just fetch the `totatSupply` at the checkpoint directly from token.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -103,6 +103,14 @@ contract PLCRVotingCheckpoint is Module {
         _validValueCheck(_commitDuration);
         _validValueCheck(_revealDuration);
         _validValueCheck(_proposedQuorum);
+        // Overflow check
+        require(
+            uint64(_commitDuration) == _commitDuration && 
+            uint64(_revealDuration) == _revealDuration &&
+            uint64(_startTime) == _startTime &&
+            uint24(_totalProposals) == _totalProposals,
+            "Uint64 values get overflowed"
```
maxsam4: Not all values are `Uint64` so maybe we should use some generic word.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,321 @@
...
+            isVotingSucceed = true;
+            for (i = 0; i < ballot.totalProposals; i++) {
+                if (maxWeight == ballot.weightedVote[i] && i != winningProposal)
+                    counter ++;
+            }
+        }
+        
+        uint256[] memory tieWith = new uint256[](counter);
+        counter = 0;
+        if (counter > 1) {
```
maxsam4: This will never be true because the counter is being set to 0 in the above line.
the counter should be set to 0 in the next line and not before. 
Also, the statement should be `counter >= 1` instead of `counter > 1`

maxsam4 (changes requested):

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,314 @@
...
+     * @param _choiceOfProposal Proposal chossed by the voter.
+     * @param _salt used salt for hashing (unique for each user)
+     */
+    function revealVote(uint256 _ballotId, uint256 _choiceOfProposal, uint256 _salt) external {
+        _validBallotId(_ballotId);
+        require(getCurrentBallotStage(_ballotId) == Stage.REVEAL, "Not in reveal stage");
+        Ballot storage ballot = ballots[_ballotId];
+        require(ballot.isActive, "Inactive ballot");
+        require(ballot.proposalToVote[msg.sender].secretVote != bytes32(0), "Should be commit vote first");
+        require(_choiceOfProposal < ballot.totalProposals, "Invalid proposal choice");
```
maxsam4: If proposals start with id 1 then we should use `<=` instead of `<` here. If proposals start with id 0 (as they should), we should make it clear here. While creating a proposal, it takes the number of proposals as input and minimum is set to two. That suggests that valid proposals are 1 and 2 rather than 0 and 1.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,314 @@
...
+     * @param _ballotId Given ballot Id
+     * @param _choiceOfProposal Proposal chossed by the voter.
+     * @param _salt used salt for hashing (unique for each user)
+     */
+    function revealVote(uint256 _ballotId, uint256 _choiceOfProposal, uint256 _salt) external {
+        _validBallotId(_ballotId);
+        require(getCurrentBallotStage(_ballotId) == Stage.REVEAL, "Not in reveal stage");
+        Ballot storage ballot = ballots[_ballotId];
+        require(ballot.isActive, "Inactive ballot");
+        require(ballot.proposalToVote[msg.sender].secretVote != bytes32(0), "Should be commit vote first");
```
maxsam4:
```suggestion
        require(ballot.proposalToVote[msg.sender].secretVote != bytes32(0), "Secret vote not available");
```
Or use something similar that covers both case: when the user didn't cast their vote or already revealed their vote.

contracts/modules/Experimental/Checkpoints/PLCRVotingCheckpoint.sol
satyamakgec: I think only need to add the comment to clear it. As we are starting the proposal Id with 0 here in this module

maxsam4 (approved):

PR #593, opened 2019-03-12
1410 Integration

adamdossa:


contracts/modules/Experimental/TransferManager/LTM/LockUpTransferManager.sol
```diff
@@ -658,20 +658,22 @@ contract LockUpTransferManager is LockUpTransferManagerStorage, TransferManager
...
-        uint256 _currentBalance = IERC20(securityToken).balanceOf(_owner);
-        if (_partition == LOCKED) {
-            return Math.min(getLockedTokenToUser(_owner), _currentBalance);
-        } else if (_partition == UNLOCKED) {
-            if (_currentBalance < getLockedTokenToUser(_owner)) {
-                return 0;
-            }
-            return _currentBalance.sub(getLockedTokenToUser(_owner));
+    function getTokensByPartition(bytes32 _partition, address _tokenHolder, uint256 _additionalBalance) external view returns(uint256){
+        uint256 currentBalance = (msg.sender == securityToken) ? (IERC20(securityToken).balanceOf(_tokenHolder)).add(_additionalBalance) : uint256(0);
```
adamdossa: Curious as to why the function has to work only if the caller is the ST? (I agree that this will be when it is actually called, but just wondering why we had to enforce this).

contracts/tokens/SecurityToken.sol
```diff
@@ -84,6 +85,17 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
             require(_isModule(msg.sender, _type));
     }
 
+    function _isValidPartition(bytes32 _partition) internal pure {
+        require(_partition == UNLOCKED, "Invalid partition");
```
adamdossa, line 112: Can it not also be LOCKED?

contracts/tokens/SecurityToken.sol
```diff
@@ -394,6 +411,143 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
...
+    
+    /**
+     * @notice Authorises an operator for all partitions of `msg.sender`.
+     * NB - Allowing investors to authorize an investor to be an operator of all partitions
+     * but it doesn't mean we operator is allowed to transfer the LOCKED partition values.
+     * Logic for this restriction is written in `operatorTransferByPartition()` function.
+     * @param _operator An address which is being authorised.
+     */ 
+    function authorizeOperator(address _operator) external {
+        approvals[msg.sender][_operator] = true;
```
adamdossa: Possible alternative approach I've seen which could maybe avoid having a separate `approvals` value is to use the normal ERC20 `allowances` variable, but set the allowance to uint(-1) and then don't subtract the amount in transferFrom if the allowance is uint(-1).

WETH uses this approach:
https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code

contracts/tokens/SecurityToken.sol
```diff
@@ -394,6 +411,143 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
...
+    }
+
+    /**
+     * @notice Authorises an operator for a given partition of `msg.sender`
+     * @param _partition The partition to which the operator is authorised
+     * @param _operator An address which is being authorised
+     */
+    function authorizeOperatorByPartition(bytes32 _partition, address _operator) external {
+        _isValidPartition(_partition);
+        partitionApprovals[msg.sender][_partition][_operator] = true;
```
adamdossa, line 569: If we are only allowing people to authorize others on the UNLOCKED partition, do we need to distinguish this at all from `authorizeOperator`?

contracts/tokens/SecurityToken.sol
adamdossa: Ignore comment - I see you're only using this for transfers / operators etc., you can still call balanceOfPartition on LOCKED.

contracts/modules/Experimental/TransferManager/LTM/LockUpTransferManager.sol
satyamakgec: because it can be called from the modules as well. And if we allow the investor to call this function with any `_additionalvalue` then it may mislead then as per the output they get. That's why I think `_additionalvalue` will be in effect only when it is called by the securityToken.

contracts/tokens/SecurityToken.sol
satyamakgec: As per the ERC definition, it should be distinguished from the `authorizeOperatorByPartition()`. I am trying to follow the same here. An investor can authorize the operator to trade every partition by calling `authorizeOperator()` but if I try to call `operatorRedeemByPartition` & `operatorTransferByPartition` with different partition beside UNLOCKED it reverts whether I am authorized to trade that partition. 

On `authorizeOperatorByPartition()` I explicitly add the valid partition check to not waste the gas of the investor by allowing the investor to authorize the operator to trade LOCKED or other partition besides UNLOCKED. By this way, I think this function also follows the standard definition as per the ERC.

contracts/tokens/SecurityToken.sol
satyamakgec: Yes, it will work with `authorizeOperator`. That wouldn't work with `authorizeOperatorByPartition` in general but for our current case, it will work as we only have UNLOCKED partition to entertain. But if we want to introduce more tradable partition in the future then upgradeability will not help as we need to introduce more state variables to tackle this. So I will use uint(-1) approach for the `approvals` not for the `partitionApprovals`

satyamakgec (approved):

PR #599, opened 2019-03-14
Add acknowledgement for irrevocable actions using EIP712

maxsam4:
Owners now need to sign acknowledgement before freezing issuance or disabling controller.

### DIFF
--- a/contracts/libraries/TokenLib.sol
+++ b/contracts/libraries/TokenLib.sol
@@ -12,6 +12,24 @@ library TokenLib {
 
     using SafeMath for uint256;
 
+    struct EIP712Domain {
+        string  name;
+        uint256 chainId;
+        address verifyingContract;
+    }
+
+    struct Acknowledgment {
+        string text;
+    }
+
+    bytes32 constant EIP712DOMAIN_TYPEHASH = keccak256(
+        "EIP712Domain(string name,uint256 chainId,address verifyingContract)"
+    );
+
+    bytes32 constant ACK_TYPEHASH = keccak256(
+        "Acknowledgment(string text)"
+    );
+
     bytes32 internal constant WHITELIST = "WHITELIST";
     bytes32 internal constant INVESTORSKEY = 0xdf3a8dd24acdd05addfc6aeffef7574d2de3f844535ec91e8e0f3e45dba96731; //keccak256(abi.encodePacked("INVESTORS"))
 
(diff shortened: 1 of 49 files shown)


maxsam4:
and this breaks coverage as well :(

contracts/libraries/TokenLib.sol
```diff
@@ -12,6 +12,24 @@ library TokenLib {
 
     using SafeMath for uint256;
 
+    struct EIP712Domain {
+        string  name;
```
satyamakgec, line 16: Do we need to add the `version` here to avoid to use the same messages when the security token is upgradable?

contracts/tokens/SecurityToken.sol
```diff
@@ -485,7 +479,8 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
      * @notice Permanently freeze issuance of this security token.
      * @dev It MUST NOT be possible to increase `totalSuppy` after this function is called.
      */
-    function freezeIssuance() external isIssuanceAllowed isEnabled("freezeIssuanceAllowed") onlyOwner {
+    function freezeIssuance(bytes calldata _signature) external isIssuanceAllowed onlyOwner {
+        require(owner() == TokenLib.recoverFreezeIssuanceAckSigner(_signature), "Owner did not sign");
```
satyamakgec, line 483: the signer is always the owner? and msg.sender is also the owner. is it right or I am missing something here

contracts/libraries/TokenLib.sol
maxsam4: If the ST is upgraded, the address (verifyingContract) will change and hence this sig will become invalid.
Also, these signs can only be used by the owners themselves so there is no question of unauthorized reuse.

contracts/tokens/SecurityToken.sol
maxsam4: yes, the signer is the owner and the msg.sender. This is just an additional acknowledgement from the owner that these actions are irrevocable. Please refer to the asana task for details.

contracts/libraries/TokenLib.sol
satyamakgec: `verifyingContract` will not change if we upgrade the ST using the proxy. or it will refer to logic contract address.

maxsam4: 1) We can not upgrade ST using proxy.
2) Even If we upgrade using proxy in future, these signatures will still remain useless because these signatures are used to trigger an irrevocable action. If the action has already been triggered, it can't be triggered again anyway. The information regarding triggering is stored in the proxy so a triggered action will remain triggered.

satyamakgec (approved):

adamdossa (approved):

test/o_security_token.js
```diff
@@ -328,37 +331,31 @@ contract("SecurityToken", async (accounts) => {
...
-            await catchRevert(I_FeatureRegistry.setFeatureStatus("freezeIssuanceAllowed", false, { from: account_polymath }));
-            assert.equal(false, await I_FeatureRegistry.getFeatureStatus("freezeIssuanceAllowed", { from: account_temp }));
-            await I_FeatureRegistry.setFeatureStatus("freezeIssuanceAllowed", true, { from: account_polymath });
-            assert.equal(true, await I_FeatureRegistry.getFeatureStatus("freezeIssuanceAllowed", { from: account_temp }));
-
-            await catchRevert(I_FeatureRegistry.setFeatureStatus("freezeIssuanceAllowed", true, { from: account_polymath }));
+        
+        it("Should finish the minting -- fail because owner didn't sign correct acknowledegement", async () => {
+            let trueButOutOfPlaceAcknowledegement = web3.utils.utf8ToHex(
+                "F O'Brien is the best!"
```
F-OBrien, line 344: lol

PR #600, opened 2019-03-14
Allow token name change

maxsam4:
Added ability to change token name
Updated OZ version


maxsam4:
#599 Reduces the ST size by a little bit so once that is merged, we should probably be able to uncomment the change name event.

contracts/tokens/SecurityToken.sol
```diff
@@ -289,6 +292,15 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
         dataStore = _dataStore;
     }
 
+    /**
+    * @notice Allows owner to change token name
+    * @param _name new name of the token
+    */
+    function changeName(string calldata _name) external onlyOwner {
```
satyamakgec, line 299: Should we add the sanity check for the `_name`? I know it only called by the owner but still, the owner can do the mistake

maxsam4: Deploying nameless ST is allowed maybe we should allow changing name to nameless as well?

What if faceless men want to deploy their own ST? (useless GOT reference)

satyamakgec: I am not sure about this. Because they need to submit the investment report to SEC so there is no chance to work as nameless. Even if they have private equity round issuer needs to come in front of its investors. 
BTW I like the reference. Winter is here !!

adamdossa: I don't see much value in having a contract level check. Even if we insisted that _name.length > 0, they could still just put something silly in (i.e. " "). The dApp can enforce some sensible scheme if needed, otherwise if the faceless men want to go wild they can do.

yarn.lock
```diff
@@ -5148,10 +5148,10 @@ onetime@^2.0.0:
   dependencies:
     mimic-fn "^1.0.0"
 
-openzeppelin-solidity@2.1.2:
-  version "2.1.2"
-  resolved "https://registry.yarnpkg.com/openzeppelin-solidity/-/openzeppelin-solidity-2.1.2.tgz#94e2bb92b60e91abb22c6fe27d983d92850fb324"
-  integrity sha512-1ggh+AZFpMAgGfgnVMQ8dwYawjD2QN4xuWkQS4FUbeUz1fnCKJpguUl2cyadyfDYjBq1XJ6MA6VkzYpTZtJMqw==
+openzeppelin-solidity@^2.2.0:
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/openzeppelin-solidity/-/openzeppelin-solidity-2.2.0.tgz#1f0e857f53bda29b77a8e72f9a629db81015fd34"
```
adamdossa, line 5153: Does this bring in any other changes to contracts we use?

maxsam4: For some reason, it reduced st size. There is no real change functionally.

adamdossa (approved):

PR #602, opened 2019-03-17
Upgradable tokens

adamdossa:
Add the ability to upgrade tokens

  - unverify a module factory when it's underlying logic contract is upgraded
  - add ability to add module in an archived state
  - remove wildcard version bounds in module factories
  - only allow module factory version bounds to be made less restrictive (lowerBound can be lowered, upperBound can be raised)
  - allow module factory, module factory owner and module registry owner to unverify a module factory
  - add `getFactoryDetails` to MR to get verified status and reputation of a module factory
  - remove `setProtocolVersion` in STR and replace with `setProtocolFactory` and `setLatestVersion` to allow STFactories to be set independently of default versions
  - check all (non-archived) modules are valid (version bounds) before allowing a token upgrade
  - check archived modules are valid before unarchiving them
  - add `changeTreasuryWallet`, `upgradeModule` & `upgradeToken` to ISecurityToken.sol
  - check module factory is verified and within bounds when upgrading modules
  - modify version bounds of current module factories to be 3.0.0 - 3.0.0

Upgrading Approach:
  - all in-place upgrades (modules or tokens) *must* be backwards compatible (minor & patch)
  - STFactory version in STR should reflect only the major version (not enforced)


contracts/ModuleRegistry.sol
```diff
@@ -118,29 +120,36 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
...
+    function useModule(address _moduleFactory, bool _isUpgrade) external {
+        if (IFeatureRegistry(getAddressValue(Encoder.getKey("featureRegistry"))).getFeatureStatus("customModulesAllowed")) {
+            require(
+                getBoolValue(Encoder.getKey("verified", _moduleFactory)) || IOwnable(_moduleFactory).owner() == IOwnable(msg.sender).owner(),
+                "ModuleFactory must be verified or SecurityToken owner must be ModuleFactory owner"
+            );
+        } else {
+            require(getBoolValue(Encoder.getKey("verified", _moduleFactory)), "ModuleFactory must be verified");
+        }
+        require(isCompatibleModule(_moduleFactory, msg.sender), "Incompatible versions");
```
maxsam4: Moving this require statement outside the `isSecurityToken` check means any user can make MR call any malicious contract. I don't currently see anything that can be exploited by this but this opens up another attack vector and to be safe, we should only allow SecurityTokens to make this call.

contracts/SecurityTokenRegistry.sol
```diff
@@ -194,7 +194,8 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
         set(PAUSED, false);
         set(OWNER, _owner);
         set(POLYMATHREGISTRY, _polymathRegistry);
-        _setProtocolVersion(_STFactory, uint8(2), uint8(0), uint8(0));
+        _setProtocolFactory(_STFactory, uint8(2), uint8(0), uint8(0));
```
maxsam4: NB As we will be in-place upgrading STR, we can not call this function because STR is already initialized so we will have to call `setProtocolFactory` and `setLatestVersion` manually. It might be best to just comment out this function to reduce code size and remind us that this can not be called when upgrading via in-place upgrade.

contracts/SecurityTokenRegistry.sol
```diff
@@ -682,34 +683,49 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
     * @param _minor Minor version of the proxy.
     * @param _patch Patch version of the proxy
     */
-    function setProtocolVersion(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) external onlyOwner {
+    function setProtocolFactory(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) external onlyOwner {
+        _setProtocolFactory(_STFactoryAddress, _major, _minor, _patch);
+    }
+
+    function _setProtocolFactory(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) internal {
         require(_STFactoryAddress != address(0), "Bad address");
```
maxsam4, line 730: Maybe we should allow `address(0)` to disable creation of security tokens of that version. Can you please confirm if we are planning on ever disabling creation of older ST versions and if there is some other way to do it?
ps. We can keep the check and use 0xdead or something similar to disable st creation to get the best of both worlds.

contracts/SecurityTokenRegistry.sol
```diff
@@ -682,34 +683,49 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
-        _setProtocolVersion(_STFactoryAddress, _major, _minor, _patch);
+        uint8[] memory _version = new uint8[](3);
```
maxsam4: I think we don't need `_version` here.

contracts/SecurityTokenRegistry.sol
```diff
@@ -682,34 +683,49 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
+    * @param _minor Minor version of the proxy.
+    * @param _patch Patch version of the proxy
     */
-    function _setProtocolVersion(address _STFactoryAddress, uint8 _major, uint8 _minor, uint8 _patch) internal {
+    function setLatestVersion(uint8 _major, uint8 _minor, uint8 _patch) external onlyOwner {
+        _setLatestVersion(_major, _minor, _patch);
+    }
+
+    function _setLatestVersion(uint8 _major, uint8 _minor, uint8 _patch) internal {
         uint8[] memory _version = new uint8[](3);
```
maxsam4: We also don't need `_version` here anymore.

contracts/tokens/STFactory.sol
```diff
@@ -31,24 +65,94 @@ contract STFactory is ISTFactory {
...
+        require(tokenToRegistry[msg.sender] != address(0), "Invalid token");
+        uint256 newVersion = tokenVersion[msg.sender] + 1;
+        require(newVersion <= latestUpgrade, "Incorrect version");
+        OwnedUpgradeabilityProxy(address(uint160(msg.sender))).upgradeToAndCall(logicContracts[newVersion].version, logicContracts[newVersion].logicContract, logicContracts[newVersion].upgradeData);
+        tokenVersion[msg.sender] = newVersion;
+        // Check that all modules remain valid
+        IModuleRegistry moduleRegistry = IModuleRegistry(IPolymathRegistry(tokenToRegistry[msg.sender]).getAddress("ModuleRegistry"));
+        address moduleFactory;
+        bool isArchived;
+        for (uint8 i = 1; i < 10; i++) {
```
maxsam4: 10 should be a customizable variable in case we decide to add more than 10 types of modules in future.

contracts/tokens/STFactory.sol
```diff
@@ -31,24 +65,94 @@ contract STFactory is ISTFactory {
...
+     */
+    function upgradeToken() external {
+        // Check the token was created by this factory
+        require(tokenToRegistry[msg.sender] != address(0), "Invalid token");
+        uint256 newVersion = tokenVersion[msg.sender] + 1;
+        require(newVersion <= latestUpgrade, "Incorrect version");
+        OwnedUpgradeabilityProxy(address(uint160(msg.sender))).upgradeToAndCall(logicContracts[newVersion].version, logicContracts[newVersion].logicContract, logicContracts[newVersion].upgradeData);
+        tokenVersion[msg.sender] = newVersion;
+        // Check that all modules remain valid
+        IModuleRegistry moduleRegistry = IModuleRegistry(IPolymathRegistry(tokenToRegistry[msg.sender]).getAddress("ModuleRegistry"));
```
maxsam4: This can make STFactory call a malicious contract. Not harmful afaict but another attack vector.

contracts/tokens/STFactory.sol
```diff
@@ -1,21 +1,55 @@
...
+import "../interfaces/IPolymathRegistry.sol";
 import "../datastore/DataStoreFactory.sol";
 
 /**
  * @title Proxy for deploying SecurityToken instances
  */
-contract STFactory is ISTFactory {
+contract STFactory is ISTFactory, Ownable {
+
     address public transferManagerFactory;
```
maxsam4, line 19: As logic contract can be upgraded, We should allow upgrading of transferManagerFactory as well?

contracts/tokens/STFactory.sol
```diff
@@ -1,21 +1,55 @@
...
-    address public stDelegate;
     DataStoreFactory public dataStoreFactory;
```
maxsam4, line 20: As logic contract can be upgraded, We should allow upgrading of dataStoreFactory as well?
Maybe even make a struct of all these address.

contracts/tokens/STFactory.sol
```diff
@@ -31,24 +65,94 @@ contract STFactory is ISTFactory {
...
+        SecurityTokenProxy proxy = new SecurityTokenProxy(
+            _name,
+            _symbol,
+            _decimals,
+            _divisible ? 1 : uint256(10) ** _decimals,
+            _tokenDetails,
+            _polymathRegistry
+        );
+        // Sets logic contract
+        proxy.upgradeTo(logicContracts[latestUpgrade].version, logicContracts[latestUpgrade].logicContract);
```
maxsam4, line 115: STF should deploy genesis version of ST rather than the latest version.

contracts/tokens/SecurityToken.sol
```diff
@@ -781,4 +813,49 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
...
+        return msg.sender == _owner;
+    }
+
+    /**
+     * @dev Allows the current owner to relinquish control of the contract.
+     * @notice Renouncing to ownership will leave the contract without an owner.
+     * It will not be possible to call the functions with the `onlyOwner`
+     * modifier anymore.
+     */
+    function renounceOwnership() public onlyOwner {
```
maxsam4: This is too risky to exist without any confirmation. We have requirement of confirmation for even freezing issuance.

contracts/tokens/STGetter.sol
```diff
@@ -11,29 +11,25 @@ contract STGetter is OZStorage, SecurityTokenStorage {
...
+     * @notice constructor
+     * @dev Never called as contract is created through Proxy factory
+     * @dev Needed otherwise contract is considered abstract
+     * @param _name Name of the SecurityToken
+     * @param _symbol Symbol of the Token
+     * @param _decimals Decimals for the securityToken
      */
-    function getAllDocuments() external view returns (bytes32[] memory) {
-        return _docNames;
+    constructor(
```
maxsam4: Why do we need to initialize this?

contracts/tokens/SecurityToken.sol
```diff
@@ -67,6 +68,40 @@ contract SecurityToken is ERC20, ERC20Detailed, Ownable, ReentrancyGuard, Securi
...
+    event TokenUpgraded(uint8 _major, uint8 _minor, uint8 _patch);
+    /**
+     * @notice constructor
+     * @dev Never called as contract is created through Proxy factory
+     * @dev Needed otherwise contract is considered abstract
+     * @param _name Name of the SecurityToken
+     * @param _symbol Symbol of the Token
+     * @param _decimals Decimals for the securityToken
+     */
+    constructor(
```
maxsam4: Why do we need to initialize this?

contracts/modules/UpgradableModuleFactory.sol
```diff
@@ -81,10 +85,14 @@ contract UpgradableModuleFactory is ModuleFactory {
         require(moduleToSecurityToken[_module] == msg.sender, "Incorrect caller");
         // Only allow issuers to upgrade in single step verisons to preserve upgradeToAndCall semantics
         uint256 newVersion = modules[msg.sender][_module] + 1;
-        /* uint256 newVersion = moduleVersions[msg.sender] + 1; */
-        require(newVersion <= latestVersion, "Incorrect version");
-        OwnedUpgradeabilityProxy(address(uint160(_module))).upgradeToAndCall(logicContracts[newVersion].version, logicContracts[newVersion].logicContract, logicContracts[newVersion].logicData);
+        require(newVersion <= latestUpgrade, "Incorrect version");
+        OwnedUpgradeabilityProxy(address(uint160(_module))).upgradeToAndCall(logicContracts[newVersion].version, logicContracts[newVersion].logicContract, logicContracts[newVersion].upgradeData);
```
maxsam4, line 91: Why do we need to cast `_module` to `address payable`?

maxsam4 (commented):
We really need an Instruction sheet for our future self to remind us of the exact order/procedure of upgrading the protocol version, module version, token etc.

contracts/ModuleRegistry.sol
adamdossa: Yep - I did think about this and the MR has no particular permissions I believe, but for safety sake (and possibly future bugs) have moved it into the if clause.

contracts/SecurityTokenRegistry.sol
adamdossa: Agreed. Ultimately I'd expect us to never call this function again. If we needed to, in a future upgrade of the STR, do some more "initialization" it should be via a new function which is called during upgradeToAndCall.

contracts/SecurityTokenRegistry.sol
adamdossa: True. Me doing bad copy / paste.

contracts/SecurityTokenRegistry.sol
adamdossa: Agreed

contracts/tokens/STFactory.sol
adamdossa: Hmm - maybe better to take the PMR as a constructor param and fix it - will look at this in more detail - leaving unresolved for now.

contracts/tokens/STFactory.sol
adamdossa: Done (if by upgrading you mean modifying the address).

contracts/tokens/STFactory.sol
adamdossa: Done (if by upgrading you mean modifying the address). Didn't bother with the struct though.

contracts/tokens/STFactory.sol
adamdossa: Don't agree - otherwise newly generated tokens will be on old versions. I think everyone should be forced to use the latest minor / patch version of a major version. Major version can be selected through the STR.

contracts/tokens/SecurityToken.sol
adamdossa: Agreed - removed. I guess this is in our current version as we inherit from Ownable which has this (unless this function isn't in the previous version of OZ),

maxsam4: Yeah, It's part of the current ST. We missed it or didn't think that it will be an Issue. The recent acknowledgement change basically made me catch this lol. I would have probably not commented anything If we hadn't made that change. 

The PR is merged now so you can just use the functions from there and change the ack text.

contracts/ModuleRegistry.sol
adamdossa: Actually - unresolving this one - moving this check inside the if clause means that module bounds are not checked when we add the GTM from the STF, as the token has not yet been registered in the STR. I can't see anything malicious that could happen from forcing the MTR to call out to an arbitrary contract but equally a STF should never be setup in such a way as to try and include an incompatible module. Neither ideal - maybe this area can be re-worked a bit at a higher level to improve the logic.

contracts/SecurityTokenRegistry.sol
adamdossa: I added: removeProtocolFactory to explicitly remove an existing protocol factory, provided you're not removing the current latestVersion.

contracts/tokens/STFactory.sol
adamdossa: Agree - this is added.

contracts/tokens/STGetter.sol
adamdossa: No - this is fixed.

contracts/tokens/SecurityToken.sol
adamdossa: ST inherits from ERC20Detailed, so we can't create the logic contract without calling the constructor of ERC20Detailed. We could refactor this if we wanted to, but probably not worth it IMO.

contracts/modules/UpgradableModuleFactory.sol
adamdossa: Agreed - not needed.

adamdossa: Actually sorry - we do need to cast to avoid:
TypeError: Explicit type conversion not allowed from non-payable "address" to "contract OwnedUpgradeabilityProxy", which has a payable fallback function.

contracts/tokens/OZStorage.sol
```diff
@@ -2,17 +2,15 @@ pragma solidity ^0.5.0;
 
 contract OZStorage {
 
-    mapping (address => uint256) private _balances;
+    mapping (address => uint256) public _balances;
```
maxsam4, line 5: public variables generate a getter with the same name that takes some code size. We should instead define these variables as internal.

contracts/ModuleRegistry.sol
```diff
@@ -221,12 +230,29 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
...
+        emit ModuleVerified(_moduleFactory);
+    }
+
+    /**
+    * @notice Called by Polymath to verify Module Factories for SecurityTokens to use.
+    * @notice A module can not be used by an ST unless first approved/verified by Polymath
+    * @notice (The only exception to this is that the author of the module is the owner of the ST)
+    * @notice -> Only if Polymath enabled the feature.
+    * @param _moduleFactory is the address of the module factory to be verified
+    */
```
satyamakgec, line 249: I think need to change the natspec comments

contracts/SecurityTokenRegistry.sol
```diff
@@ -184,17 +184,15 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
     {
         require(!getBoolValue(INITIALIZE),"Initialized");
         require(
-            _STFactory != address(0) && _owner != address(0) && _polymathRegistry != address(0) && _getterContract != address(0),
+            _owner != address(0) && _polymathRegistry != address(0) && _getterContract != address(0),
             "Invalid address"
         );
-        require(_stLaunchFee != 0 && _tickerRegFee != 0, "Bad fee");
```
satyamakgec, line 193: any particular reason to remove this check?

contracts/tokens/OZStorage.sol
```diff
@@ -2,17 +2,15 @@ pragma solidity ^0.5.0;
...
 
-    uint256 private _totalSupply;
+    uint256 public _totalSupply;
 
-    string private _name;
-    string private _symbol;
-    uint8 private _decimals;
-
-    address private _owner;
+    string public _name;
```
satyamakgec: use internal instead of public as `_name` or other variables following the internal functions/variable standard. and using the `public` also affect the transparency of the proxy

migrations/2_deploy_contracts.js
```diff
@@ -382,7 +398,6 @@ module.exports = function(deployer, network, accounts) {
         .then((securityTokenRegistryProxy) => {
```
satyamakgec, line 398: Shouldn't we call `setProtocolFactory()` and `setLatestVersion()` in the migrations

satyamakgec (dismissed):
Only concerns are unsafe calls in STFactory. but I read the comments @adamdossa keeping those unresolved (or not exploiting anything particular).

contracts/ModuleRegistry.sol
adamdossa: I put this logic back inside the `if` clause. I also perform the check (compatible modules) if the caller is the current STFactory. This still leaves a gap if the caller is an old STFactory (i.e. someone has used generateST passing in an old protocol version). This could be solved if we add something like: isSTFactory(address) into the STR, but would need the STR to maintain a list of active STFactories. Will add an Asana for this but not covering in this PR.

contracts/tokens/STFactory.sol
adamdossa: I've changed this to take the PMR as a constructor parameter in the STFactory. This means we don't need to pass the PMR as a parameter to deployToken anymore, but had to leave this in for backwards compatibility with previous STFactories.

contracts/tokens/OZStorage.sol
adamdossa: But we'll lose the etherscan.io read contract tab with proxied tokens, so having basic information available seems useful in etherscan.

contracts/ModuleRegistry.sol
adamdossa: Fixed

contracts/SecurityTokenRegistry.sol
adamdossa: Well - we allow zero fees when changing fees, and seemed a bit arbitrary to not allow it here.

contracts/tokens/OZStorage.sol
adamdossa: I flipped these from internal to public so that etherscan.io would show them. You're correct it could result in a transparency clash but seems a good reason to keep them as "_..." as we shouldn't have any public / external functions with this name schema. If you think this is a bad idea please unresolve and ping me to discuss.

migrations/2_deploy_contracts.js
adamdossa: Done

contracts/tokens/OZStorage.sol
satyamakgec: Hmm... I agree to show on etherscan we have to compromise something. If we are using those only to show the name/symbol on token tracker then we can do something like SpiceVC token.
[SpiceVC token](https://etherscan.io/address/0x0324dd195d0cd53f9f07bee6a48ee7a20bad738f) is not having the functionality of reading name & symbol from the contract itself using `read contract` but on the etherscan token tracker they are showing the token name/symbol. I think they explicitly add those stuff on the etherscan. May we do the same for our STs instead allowing them to read directly from the contract.

If we are trying to show the creadibility/proof of the contract (whether is the right contract or not) then yes I am agree we can go with your approach.

contracts/ModuleRegistry.sol
```diff
@@ -129,12 +130,16 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
         } else {
             require(getBoolValue(Encoder.getKey("verified", _moduleFactory)), "ModuleFactory must be verified");
         }
-        require(isCompatibleModule(_moduleFactory, msg.sender), "Incompatible versions");
         // This if statement is required to be able to add modules from the STFactory contract during deployment
         // before the token has been registered to the STR.
-        if ((!_isUpgrade) && ISecurityTokenRegistry(getAddressValue(Encoder.getKey("securityTokenRegistry"))).isSecurityToken(msg.sender)) {
-            pushArray(Encoder.getKey("reputation", _moduleFactory), msg.sender);
-            emit ModuleUsed(_moduleFactory, msg.sender);
+        if (ISecurityTokenRegistry(getAddressValue(Encoder.getKey("securityTokenRegistry"))).isSecurityToken(msg.sender)
```
maxsam4: To reduce gas, we can save STR address locally and then reuse it.

contracts/ModuleRegistry.sol
maxsam4: >I also perform the check (compatible modules) if the caller is the current STFactory

Why do we need this check? When STF calls addModule on the security token to add the GTM, the security token will call Module Factory to verify bounds so the msg.sender that MR gets will be of the new (unregistered) ST instead of STF. Marking this as unresolved as I couldn't understand the reasoning.

contracts/tokens/OZStorage.sol
maxsam4: >seems a good reason to keep them as "_..." as we shouldn't have any public / external functions with this name schema.

Clash can still happen. Solidity only stores the function selectors. Function selectors are first 8 digits of the hash of function signatures. There is a relatively high chance of having two completely different functions with the first 8 digits of the hash same. We have the same issue with ST and STGetter, STR and STRGetter (mentioned in my audit :)) but they aren't as major as the function that will be overshadowed in their case will be a getter function. Here, a getter function can overshadow a critical function. It's highly unlikely that this is happening right now or will happen in future when we add more functions but this is a risk I don't think we should be taking. 

I understand that it will be good for etherscan so may I suggest that we add a pre-commit hook/Travis script that will calculate function hashes of all the functions and verify that there are no clashes? Perhaps we can do it in a different PR but it should be part of the 3.0 release. Marking this as unresolved for now.

contracts/ModuleRegistry.sol
adamdossa: You are correct - the code makes no difference as msg.sender is the ST, not the STF - my mistake - will remove this line again, and we are back in the (current) situation that STF module deployments don't check bounds. We just need to be careful about making sure that STF's are "self-consistent".

contracts/tokens/OZStorage.sol
adamdossa: Agree with your approach. A pre-commit scan for any clashing sigs across our non-transparent contracts would be useful. Could probably be widened to look for any sig. clashes across any of our modules (as any sig clash could be problematic if the dApp sent a tx to the wrong contract). I am going to create an Asana / JIRA task for this, then resolve this issue.

adamdossa: https://polymath.atlassian.net/browse/SC-1

contracts/ModuleRegistry.sol
adamdossa: This is only referenced in one place in this function now.

maxsam4 (approved):

PR #558, opened 2019-02-19
POLY Only Capped STO

F-OBrien:
I've marked this as WIP as some changes will be required when some of the current PR's are merged but it is ready for review if you have time. (hopefully before too many more changes are needed to keep up with you guys :smile:)

As I've mentioned before I'm not a developer but took an interest in learning solidity so gave myself this project which I think could be of value to Polymath as an open source contribution. I'm open to any comments, optimizations, change requests, criticism etc. 

### Please check if the PR fulfills these requirements
- [x] The commit message follows our Submission guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

### What kind of change does this PR introduce? 
This PR includes a new capped STO module "POLYCappedSTO"
It has been update to a proxy deployment and is in line with the latest changes in dev-3.0.0. This STO module is based on a combination of the existing CappedSTO and USDTieredSTO.

Issuer can set a hard Cap in number of tokens, POLY to token rate, start time and end time as per the CappedSTO. This module is not tiered and only accepts POLY as the Fund Raise type so can be priced separately from modules that allow other funding options.
Additional features include:
- A minimum Investment limit
- The STO supports accredited and non accredited investors and options to override the default limits for non accredited investors.
- A investment limit for non-accredited investors
- A setting maximum number of non-accredited investors
- Setting maximum number of non-accredited investors to 0 disables the limit. (Should an issuer want to disable non-accredited investors it can be done by no whitelisting them or the default non accredited  limit can be set to 0 or less than the minimum investment limit)
- Option to finalize the STO at any time.
- Option to mint or not unsold tokens upon finalization.
- Records investment values in USD at the time of of investment for individual investors and total investment.
- Allows modification to all STO configuration parameters prior to commencement of the STO.
- Supports divisible and non-divisible tokens
- Includes a pre-purchase check function to allow the investor to check the number of tokens they will receive and amount they will spend based on limits associated with their address.

### What is the current behavior? 
N/A
### What is the new behavior?
See above
### Does this PR introduce a breaking change?
No
### Any Other information:
I did some work on a CLI for this but recent commits will have broken it and as the CLI's need to be updated for 3.0.0 changes any way I've removed to avoid merge conflicts.

PR #603, opened 2019-03-17
POLY Pegged STO

F-OBrien:
This STO contract builds on the POLYCappedSTO I submitted in PR #558. It similarly only accepts POLY but it pegs the token price to the USD value of POLY at the time of investment, like the USDTieredSTO, instead of a fixed POLY rate.


contracts/modules/Experimental/STO/POLYPeggedSTO.sol
```diff
@@ -0,0 +1,572 @@
...
+     */
+    function configure(
+        uint256 _startTime,
+        uint256 _endTime,
+        uint256 _cap,
+        uint256 _rate,
+        uint256 _minimumInvestment,
+        uint256 _nonAccreditedLimit,
+        uint256 _maxNonAccreditedInvestors,
+        address payable _wallet,
```
satyamakgec, line 95: no need of payable here, as _wallet no longer receive ETHs

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
```diff
@@ -0,0 +1,572 @@
...
+        uint256 _maxNonAccreditedInvestors
+    ) internal {
+        minimumInvestment = _minimumInvestment;
+        nonAccreditedLimit = _nonAccreditedLimit;
+        maxNonAccreditedInvestors = _maxNonAccreditedInvestors;
+        emit SetLimits(minimumInvestment, nonAccreditedLimit, maxNonAccreditedInvestors);
+    }
+
+    function _modifyRate(uint256 _rate) internal {
+        require(_rate > 0, "Rate not > 0");
```
satyamakgec: Try to write the reason strings in a way that the developer understands what went wrong. Not what will be the valid value for that variable. (If there is a size issue then it can be considered).

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
```diff
@@ -0,0 +1,572 @@
...
+     * @notice Finalizes the STO and mints remaining tokens to treasury address
+     * @notice Treasury address must be whitelisted to successfully finalize
+     * @param _mintUnsoldTokens unsold tokens will be minted to the Treasury wallet if ture
+     */
+    function finalize(bool _mintUnsoldTokens) external {
+        _onlySecurityTokenOwner();
+        require(!isFinalized, "Already finalized");
+        isFinalized = true;
+        if ((_mintUnsoldTokens) && (totalTokensSold < cap)) {
+            address treasury = (treasuryWallet == address(0) ? IDataStore(getDataStore()).getAddress(TREASURY) : treasuryWallet);
```
satyamakgec, line 230: Add one more check for treasury wallet to validate `treasury != 0x0`. It will act as a one more safety measure if something bad happens with dataStore storage.

satyamakgec (commented):
@F-OBrien code looks good, minor improvements comment

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
F-OBrien: As wallet is defined as payable in STOstorage.sol if I remove payable I get an error when compiling. I have left it as payable so it is using a consistent variable across STO's but could change it to a new variable if preferred. 
"Type address is not implicitly convertible to expected type address payable.                                                                                                                                                                                                                              wallet = _wallet;
Compilation failed. See above."

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
F-OBrien: I've changed a few. Let me know if there is a specific one that you'd like changed. I've them here for your convenience:
- STO already started
- Already configured
- Invalid rate
- Invalid cap
- Invalid times
- Invalid wallet
- Already finalized
- Invalid treasury address
- Length mismatch
- Does not change value
- Beneficiary is not funder
- Unauthorized beneficiary
- Transfer failed
- STO not open
- No funds sent
- Invalid beneficiary
- Less than minimum investment
- Investment limit reached
- Max non-accredited investors
- Insufficient tokens minted
- Invalid oracle
- No tokens

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
satyamakgec: Yeah, it makes sense it will not allow you to do that if it is already declared. No need to declare a new variable you can keep as it is. 
FYI - payable address work as non-payable address but the reverse is not true. To work it around you could use `address payable add = uint160(non-payable address)`.

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
satyamakgec: You could change this `Does not change value` to be more specific.

F-OBrien: It's hard to make this more specific and and try to keep the length short with the variable name being 27 characters long. Would you prefer `Does not change allowBeneficialInvestments` or `allowBeneficialInvestments unchanged` which both exceed the length of a 32 byte string? 

Given that the function `changeAllowBeneficialInvestments` would typically be called as a stand alone function with only a single input variable it should be clear to a developer why it reverted so may not need to be more specific.

For reference USDTieredSTO.sol uses `Value unchanged` while CappedSTO.sol also uses `Does not change value` for the same function.

contracts/modules/Experimental/STO/POLYPeggedSTO.sol
F-OBrien: I googled this and got it to work using `wallet = address(uint160(_wallet));` Do you want me to make this change? While `wallet` won't receive ETH is this more or less clear to declare `_wallet `as payable than casting the non payable address to a payable one later?

satyamakgec: To receive the tokens minimum requirement is non-payable address. and it is possible to convert the non-payable address to payable as shown in the above comments.  As per your code, you are using payable wallet address because it is already declared in the `STOStorage.sol`. So I would prefer to keep as it is no need to convert it as it doesn't make sense to convert the payable address to non-payable because it will work same when you play with tokens.

PR #601, opened 2019-03-15
gtm matm optimizations

maxsam4:
Added code comments and optimized a few things.

PR #604, opened 2019-03-18
Gtm matm optimizations

maxsam4:
Ported changes from #601

### DIFF
--- a/contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
+++ b/contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
@@ -108,10 +108,10 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
         if (_data.length > 32) {
             address target;
             uint256 nonce;
-            uint64 validFrom;
-            uint64 validTo;
+            uint256 validFrom;
+            uint256 validTo;
             bytes memory data;
-            (target, nonce, validFrom, validTo, data) = abi.decode(_data, (address, uint256, uint64, uint64, bytes));
+            (target, nonce, validFrom, validTo, data) = abi.decode(_data, (address, uint256, uint256, uint256, bytes));
             if (target == address(this))
                 _processTransferSignature(nonce, validFrom, validTo, data);
         }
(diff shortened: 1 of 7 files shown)


contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -260,57 +264,57 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+    * @param _canSendAfter is the moment when the sale lockup period ends and the investor can freely sell his tokens
+    * @param _canReceiveAfter is the moment when the purchase lockup period ends and the investor can freely purchase tokens from others
     * @param _expiryTime is the moment till investors KYC will be validated. After that investor need to do re-KYC
     */
     function modifyKYCDataMulti(
         address[] memory _investors,
-        uint256[] memory _fromTime,
-        uint256[] memory _toTime,
-        uint256[] memory _expiryTime
+        uint64[] memory _canSendAfter,
```
satyamakgec, line 303: in 2.1.0 you used `uint256` in function arguments and then typecast those into uint64 in the function body. Shouldn't we make consistency in both branches

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -18,19 +18,19 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+    // if allowAllWhitelistIssuances is TRUE, then _canReceiveAfter is ignored when receiving tokens from the issuance address
+    // if allowAllWhitelistTransfers is TRUE, then _canReceiveAfter and _canSendAfter is ignored when sending or receiving tokens
     // in any case, any investor sending or receiving tokens, must have a _expiryTime in the future
     event ModifyKYCData(
         address indexed _investor,
         address indexed _addedBy,
-        uint256 _fromTime,
-        uint256 _toTime,
-        uint256 _expiryTime
+        uint64 _canSendAfter,
```
satyamakgec, line 31: Shouldn't it uint256 for consistency with dev-2.1.0

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -260,57 +264,57 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+    * @param _canReceiveAfter is the moment when the purchase lockup period ends and the investor can freely purchase or receive tokens from others
     * @param _expiryTime is the moment till investors KYC will be validated. After that investor need to do re-KYC
     */
     function modifyKYCData(
         address _investor,
-        uint256 _fromTime,
-        uint256 _toTime,
-        uint256 _expiryTime
+        uint64 _canSendAfter,
+        uint64 _canReceiveAfter,
```
satyamakgec, line 274: should be uint256

satyamakgec (changes requested):

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
maxsam4: We used `uint256` in 2.1 to make the interface compatible with 2.0. (We want to use the same dApp on 2.0 and 2.1 version of security token/modules)

There is no need to keep 3.0 and 2.1 compatible as they will require different dApps anyway.

satyamakgec (approved):

PR #610, opened 2019-03-19
Port dividend fix

satyamakgec:
port the fixes from #606


satyamakgec:
This only contains the porting of fix not adding the reclaimETH/reclaimERC20 in every module. working on it.

contracts/modules/Checkpoint/DividendCheckpoint.sol
```diff
@@ -265,6 +266,25 @@ contract DividendCheckpoint is DividendCheckpointStorage, ICheckpoint, Module {
...
+     * @dev will effectively refreeze claims on that dividend until the new maturity date passes
+     * @ dev NB - setting the expiry date to a past date will mean no more payments can be pulled
+     * @dev or pushed out of a dividend
+     * @param _dividendIndex Dividend to withdraw from
+     * @param _maturity updated maturity date
+     * @param _expiry updated expiry date
+     */
+    function updateDividendDates(uint256 _dividendIndex, uint256 _maturity, uint256 _expiry) external withPerm(ADMIN) {
+        require(_dividendIndex < dividends.length, "Invalid dividend");
+        require(_expiry > _maturity, "Expiry before maturity");
```
maxsam4, line 281: Maybe we should add a check to prevent increasing `_maturity` and `_expiry` date if the dividend has already expired.

contracts/modules/STO/STO.sol
```diff
@@ -45,20 +31,12 @@ contract STO is ISTO, STOStorage, Module, Pausable {
 
     /**
      * @notice Pause (overridden function)
+     * @dev Only securityToken owner restriction applied on the super function
      */
     function pause() public {
```
maxsam4, line 36: `module.sol` already has this function, why do we need it here?

satyamakgec: This is because we have a special condition in pause function for the STO.

contracts/modules/Checkpoint/DividendCheckpoint.sol
satyamakgec: Yeah, that could be done.

satyamakgec:
@maxsam4 could you revisit this PR. suggested changes are done

maxsam4 (approved):

adamdossa (approved):

PR #616, opened 2019-03-26
Define price in poly or usd

maxsam4:
Allows to set module, ticker and st price in Poly or USD


contracts/SecurityTokenRegistry.sol
```diff
@@ -670,6 +679,25 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
+    /**
+    * @notice Sets the ticker registration and ST launch fee amount and currency
+    * @param _tickerRegFee is the ticker registration fee (base 18 decimals)
+    * @param _stLaunchFee is the st generation fee (base 18 decimals)
+    * @param _isFeeInPoly defines if the fee is in poly or usd
+    */
+    function changeFeesAmountAndCurrency(uint256 _tickerRegFee, uint256 _stLaunchFee, bool _isFeeInPoly) external onlyOwner {
+        uint256 tickerFee = getUintValue(TICKERREGFEE);
+        uint256 stFee = getUintValue(STLAUNCHFEE);
+        bool oldIsFeesInPoly = getBoolValue(IS_FEE_IN_POLY);
```
satyamakgec: `oldIsFeesInPoly` could be renamed to `isOldFeesInPoly` for clarity

contracts/SecurityTokenRegistry.sol
```diff
@@ -90,6 +91,8 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
     event ChangeSecurityLaunchFee(uint256 _oldFee, uint256 _newFee);
     // Emit when changeTickerRegistrationFee is called
     event ChangeTickerRegistrationFee(uint256 _oldFee, uint256 _newFee);
+    // Emit when Fee currency is changed
+    event ChangeFeeCurrency(bool _oldIsFeesInPoly, bool _newIsFeesInPoly);
```
satyamakgec: `is` should be prefixed in the param names.

contracts/SecurityTokenRegistry.sol
```diff
@@ -670,6 +679,25 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
+    * @param _tickerRegFee is the ticker registration fee (base 18 decimals)
+    * @param _stLaunchFee is the st generation fee (base 18 decimals)
+    * @param _isFeeInPoly defines if the fee is in poly or usd
+    */
+    function changeFeesAmountAndCurrency(uint256 _tickerRegFee, uint256 _stLaunchFee, bool _isFeeInPoly) external onlyOwner {
+        uint256 tickerFee = getUintValue(TICKERREGFEE);
+        uint256 stFee = getUintValue(STLAUNCHFEE);
+        bool oldIsFeesInPoly = getBoolValue(IS_FEE_IN_POLY);
+        require(oldIsFeesInPoly != _isFeeInPoly, "Currency unchanged");
+        emit ChangeTickerRegistrationFee(tickerFee, _tickerRegFee);
```
satyamakgec: better to use _changeTickerRegistrationFee() and _changeSecurityLaunchFee() function here

contracts/interfaces/IModuleFactory.sol
```diff
@@ -71,6 +72,14 @@ interface IModuleFactory {
      */
     function changeUsageCost(uint256 _newUsageCost) external;
 
+    /**
+     * @notice Used to change the currency and amount of usage and setup cost
+     * @param _setupCost new setup cost
+     * @param _usageCost new usage cost
+     * @param _isCostInPoly new usage cost currency. USD or POLY
+     */
+    function changeCostsAndType(uint256 _setupCost, uint256 _usageCost, bool _isCostInPoly) external;
```
satyamakgec, line 81: can we use a similar name as present in the STR contract like `changeCostsAmountAndCurrency()` for consistency?

test/b_capped_sto.js
```diff
@@ -837,6 +837,19 @@ contract("CappedSTO", async (accounts) => {
             });
         });
 
+        describe("Pricing Test cases for Module Factory", async () => {
+            it("Should return correct price when price is in poly", async () => {
+                let newFactory = await CappedSTOFactory.new(
```
satyamakgec, line 844: I don't think this test full filling its `it statement`.

contracts/SecurityTokenRegistry.sol
maxsam4: It was originally meant to convey Old "isFeesInPoly". Changed it now.

contracts/SecurityTokenRegistry.sol
maxsam4: I decided to remove `_oldIsFeesInPoly` as it is always gonna be the opposite of new fees. renamed the new fees var.

contracts/SecurityTokenRegistry.sol
maxsam4: done

contracts/interfaces/IModuleFactory.sol
maxsam4: Hmm. I tried to keep this consistent with the naming in module factory. We call them costs and types instead of fees and currency in the module factory.

test/b_capped_sto.js
maxsam4: Not sure where the asserts went lol. I'll add them back.

satyamakgec (approved):

contracts/SecurityTokenRegistry.sol
```diff
@@ -666,8 +679,29 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
+    * @notice Sets the ticker registration and ST launch fee amount and currency
+    * @param _tickerRegFee is the ticker registration fee (base 18 decimals)
+    * @param _stLaunchFee is the st generation fee (base 18 decimals)
+    * @param _isFeeInPoly defines if the fee is in poly or usd
+    */
+    function changeFeesAmountAndCurrency(uint256 _tickerRegFee, uint256 _stLaunchFee, bool _isFeeInPoly) external onlyOwner {
+        uint256 tickerFee = getUintValue(TICKERREGFEE);
+        uint256 stFee = getUintValue(STLAUNCHFEE);
+        bool isOldFeesInPoly = getBoolValue(IS_FEE_IN_POLY);
+        require(isOldFeesInPoly != _isFeeInPoly, "Currency unchanged");
```
adamdossa, line 700: So you can only use this function if you're changing from USD <> POLY fees (not if you're just changing the fee amounts)?

adamdossa (approved):

contracts/SecurityTokenRegistry.sol
maxsam4: I added this check later as a sanity check. We already have functions to change amounts, If I remove this check, there is no sanity check left and tx will go through even if the owner forgot to enter any data and calls the function with 000000...000.

If the fee is already in poly then the owner will still be allowed to enter 000000.000 so this is not a foolproof check and I am rather agnostic about keeping it. Let me know if you think we should remove the check to allow the owner to change both prices in a single transaction.

PR #628, opened 2019-04-04
Minor optimizations

maxsam4:
Collection of minor non-contentious optimizations found in the internal review.
I'll keep adding more as I do them.


satyamakgec (dismissed):

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -236,14 +247,14 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
-        uint256 _transferType,
+        TransferType _transferType,
         bool _fromValidKYC,
         bool _toValidKYC,
         bool _fromRestricted,
         bool _toRestricted
     ) internal {
-        require(_transferType < 3, "Invalid TransferType");
-        transferRequirements[_transferType] =
+        require(uint8(_transferType) < 3, "Invalid TransferType");
```
satyamakgec: It can be removed

maxsam4: why?

contracts/tokens/SecurityToken.sol
```diff
@@ -71,6 +63,15 @@ contract SecurityToken is ERC20, ReentrancyGuard, SecurityTokenStorage, IERC1594
...
 
+    // Emit when Module get archived from the securityToken
+    // event ModuleArchived(uint8[] _types, address _module); Event emitted by the tokenLib.
+    // Emit when Module get unarchived from the securityToken
+    // event ModuleUnarchived(uint8[] _types, address _module); Event emitted by the tokenLib.
+    // Emit when Module get removed from the securityToken
+    // event ModuleRemoved(uint8[] _types, address _module); Event emitted by the tokenLib.
+    // Emit when the budget allocated to a module is changed
+    // event ModuleBudgetChanged(uint8[] _moduleTypes, address _module, uint256 _oldBudget, uint256 _budget); Event emitted by the tokenLib.
+
```
satyamakgec, line 74: Why did we commented them? AFAICT these events are emitted by the `TokenLib` library and we have to also declare them in the token contract to make them visible to Etherscan. isn't?

satyamakgec (commented):
LGTM, Please have a look on the above comments

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
satyamakgec: It will be automatically checked as the `_transferType` value can't be 3 or more than 3 according to the enum declared

contracts/tokens/SecurityToken.sol
maxsam4: > we have to also declare them in the token contract to make them visible to Etherscan. isn't?

I am not sure if that is the case. Have you tested? If it is the case, I'll add these back.

satyamakgec: yes, I am pretty sure about this.

satyamakgec (approved):

adamdossa (approved):

PR #632, opened 2019-04-05
Refresh / Upgrade tokens

maxsam4:
This will allow Issuers to regenerate a fresh security token in the same ticker name if they want to do an upgrade or want to reset the data.

### DIFF
--- a/contracts/SecurityTokenRegistry.sol
+++ b/contracts/SecurityTokenRegistry.sol
@@ -553,8 +553,7 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
         require(bytes(_name).length > 0 && bytes(_ticker).length > 0, "Bad ticker");
         require(_treasuryWallet != address(0), "0x0 not allowed");
         string memory ticker = Util.upper(_ticker);
-        bytes32 statusKey = Encoder.getKey("registeredTickers_status", ticker);
-        require(getBoolValue(statusKey), "not deployed");
+        require(_tickerStatus(ticker), "not deployed");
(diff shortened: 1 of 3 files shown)


satyamakgec:
This will only work when we are not deploying the STRProxy again? @maxsam4

maxsam4:
> This will only work when we are not deploying the STRProxy again? @maxsam4

We will be upgrading the STR implementation contract rather than deploying STRProxy again.

satyamakgec (approved):

contracts/SecurityTokenRegistry.sol
```diff
@@ -525,7 +528,38 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
...
+        bool _divisible,
+        address _treasuryWallet
+    )
+        public whenNotPausedOrOwner returns (address)
+    {
+        require(bytes(_name).length > 0 && bytes(_ticker).length > 0, "Bad ticker");
+        require(_treasuryWallet != address(0), "0x0 not allowed");
+        string memory ticker = Util.upper(_ticker);
+        bytes32 statusKey = Encoder.getKey("registeredTickers_status", ticker);
+        require(getBoolValue(statusKey), "not deployed");
```
adamdossa: Could use `_tickerStatus` instead

adamdossa (approved):

contracts/SecurityTokenRegistry.sol
maxsam4: done, thanks for pointing out.

VictorVicente (approved):

PR #640, opened 2019-04-09
Increased STR compatibility

maxsam4:
Resolves https://polymath.atlassian.net/browse/SC-7


contracts/SecurityTokenRegistry.sol
```diff
@@ -555,9 +602,17 @@ contract SecurityTokenRegistry is EternalStorage, Proxy {
         _storeSecurityTokenData(newSecurityTokenAddress, _ticker, _tokenDetails, now);
         set(Encoder.getKey("tickerToSecurityToken", _ticker), newSecurityTokenAddress);
         /*solium-disable-next-line security/no-block-members*/
-        emit NewSecurityToken(
-            _ticker, _name, newSecurityTokenAddress, msg.sender, now, msg.sender, false, _usdFee, _polyFee, _protocolVersion
-        );
+        if (_protocolVersion == getUintValue(Encoder.getKey("latestVersion"))) {
```
satyamakgec: I think instead of "latestVersion" we have to use if _protocolVersion less than 3.0.0 then it will emit the old event because when we active the 3.0 (i.e latestVersion is 3.0) and deprecate the 2.0 still it always emit the older event.

maxsam4: My thinking is to NOT set 3.0 as the latestVersion until we have a dApp that supports it. 2.0 will remain latestVersion and we will just set the factory address of 3.0 in STR to allow users to create 3.0 tokens if they want.

satyamakgec: But in that case we will never able to pass 0 as a protocolVersion after 3.0 released

satyamakgec (approved):

contracts/SecurityTokenRegistry.sol
satyamakgec: Idea is to redeploy the STR again once 2.0 is deprecated.

maxsam4: I have hardcoded version checks to 2.0.0 for old and 3.0.0 for new. We can now set latestVersion to 3.0.0 without needing to upgrade STR.
However, we will need to upgrade STR logic contract when we release the next version of ST after 3.0.0.

maxsam4:
Prettified PCLR test cases as well (unintentionally :)).

maxsam4:
Unprettified the test cases to avoid merge conflicts with https://github.com/PolymathNetwork/polymath-core/pull/645

satyamakgec (approved):

PR #655, opened 2019-04-23
Restricted partial sale TM

satyamakgec:


contracts/modules/Experimental/TransferManager/RestrictedPartialSaleTM.sol
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return Result.NA;
+    }
+
+    /**
+     * @notice Add/Remove wallet address from the exempt list
+     * @param _wallet Ethereum wallet/contract address that need to be exempted
+     * @param _change Boolean value used to add (i.e true) or remove (i.e false) from the list
+     */
+    function changeExemptWalletList(address _wallet, bool _change) external withPerm(ADMIN) {
```
maxsam4: Please rename `_change` to something like `_exempted`. 
change as a name suggests that the current status should be changed (toggled).

satyamakgec: Yes, the current status should be changed. that is the whole idea. A switch to add or remove the addresses from the exemption list.

maxsam4: As status must always be changed, that means change = true always.
However, that's not what change variable represents. It represents the value to which it should be changed rather than if it should be changed or not. As it represents the value, it should be called as the value.

adamdossa:
@satyamakgec Couple of comments:
  - I didn't see the specification, but is it def. correct that we whitelist exempt addresses, rather than in-scope addresses? I'd have thought (but could be wrong) that the common case is that this restriction would apply to just a couple of investors with most investors not being restricted (which would imply it should be the other way around).
  - I think changes to exemption list should be an OPERATOR rather than ADMIN function (since it doesn't put any funds at risk)

adamdossa (approved):

PR #663, opened 2019-05-02
Add signed data under canTransfer check

satyamakgec:
`_to` address can be whitelisted and checked for the transfer.


contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -480,41 +534,71 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
 
         bytes32 hash = keccak256(
             abi.encodePacked(this, _investor, _canSendAfter, _canReceiveAfter, _expiryTime, _validFrom, _validTo, _nonce)
         );
 
-        if (_checkSig(hash, _signature, _nonce)) {
+        (isVerifiedSig, signer) = _checkSigView(hash, _signature, _nonce);
+        if (isVerifiedSig) {
+            if (_isReadOnlyCall) {
+                for (uint256 i = 0; i < _investor.length; i++) {
```
adamdossa: Doesn't this mean that if the if statements fail in L571 or L575 the function will return (true, ...) even if `_isReadOnlyCall == true`? Not sure why we're only doing the uint64 checks when `_isReadOnlyCall == true`.

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -134,34 +134,85 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
+                    bool success, 
+                    address[] memory investor,
+                    ,
+                    uint256[] memory canRecieveAfter,
+                    uint256[] memory expiryTime
+                ) = _processTransferSignatureView(data, validFrom, validTo, nonce);
+                if (success) {
+                    for (nonce = 0; nonce < investor.length; nonce++) {
+                        // Searching the _to token holder to get the required details for _verifyTransfer validation 
+                        if (investor[nonce] == _to)
```
adamdossa, line 171: This logic doesn't quite match the `transferWithData` version as it is possible the signed data is relevant for the `from` address which wouldn't be picked up here I think.

adamdossa (changes requested):
@satyamakgec it looks like there is a difference in behaviour in the GTM between `canTransferWithData` and `transferWithData`

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
satyamakgec: Do you think `from` address will come under into the check, I am expecting `from` does posses the balance it means It is already whitelisted with the ST?

adamdossa (approved):

PR #668, opened 2019-05-07
Minor transfer optimization

maxsam4:
Apart from this, in the security token, we pay about 10k extra gas for using a library. If we can fit those functions in the contract itself, we'll save that gas. That'll require significant changes though.

There's not really much else we can do in the ST itself. We can try optimizing individual transfer managers.


adamdossa (approved):

tintinweb:
👍

PR #672, opened 2019-05-24
Added constructors

maxsam4:
The constructors prevent an exploit in case we forget to initialize the implementation contracts.
Ideally, we should always remember to initialize implementation contracts as well.

Fixes audit Issue 1

ps. The same exploit was possible in STR If we forgot to initialize STR implementation. I have fixed that as well in this PR.


satyamakgec (approved):

tintinweb:
👍 tin/diligence

PR #674, opened 2019-05-24
Made onlyModuleOrOwner definition consisitent

maxsam4:
Fixes audit issue 19


adamdossa (approved):

tintinweb:
👍it is good practice and recommended to always add braces `{` `}` for conditional branches even though in that case they are not necessary and it is unlikely that another statement will be added to the `true` branch. 

tin/diligence

PR #677, opened 2019-05-24
Remove the custom module logic.

satyamakgec:
fix 3.5 & 3.7


contracts/ModuleRegistry.sol
```diff
@@ -258,9 +211,8 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
     function unverifyModule(address _moduleFactory) external {
         // Can be called by the registry owner, the module factory, or the module factory owner
         bool isOwner = msg.sender == owner();
-        bool isFactoryOwner = msg.sender == IOwnable(_moduleFactory).owner();
         bool isFactory = msg.sender == _moduleFactory;
-        require(isOwner || isFactoryOwner || isFactory, "Not authorised");
+        require(isOwner || isFactory, "Not authorised");
```
maxsam4, line 215: Maybe we should allow `isFactoryOwner` to unverify their modules as well? Given that we might register and verify custom modules for issuers while keeping them as the actual factory owner.

maxsam4:
Closing this in favor of https://github.com/PolymathNetwork/polymath-core/pull/698

PR #679, opened 2019-05-25
Fix the BalanceOfPartition audit item

satyamakgec:


adamdossa (approved):

contracts/tokens/SecurityToken.sol
```diff
@@ -487,13 +487,20 @@ contract SecurityToken is ERC20, ReentrancyGuard, SecurityTokenStorage, IERC1594
...
+    function _balanceOfByPartition(bytes32 _partition, address _tokenHolder, uint256 _additionalBalance) internal view returns(uint256 partitionBalance) {
         address[] memory tms = modules[TRANSFER_KEY];
         uint256 amount;
         for (uint256 i = 0; i < tms.length; i++) {
             amount = ITransferManager(tms[i]).getTokensByPartition(_partition, _tokenHolder, _additionalBalance);
-            if (max < amount) {
-                max = amount;
+            // In UNLOCKED partition we are returning the minimum of all the unlocked balances
+            if (_partition == UNLOCKED) {
+                if (amount < partitionBalance || i == 0)
```
tintinweb, line 497: Some transfer managers do not implement this `getTokensByPartition()` properly and just return 0 (default implementation in baseclass). This might need to be fixed by implementing `getTokensByPartition` function in every TM or adding a flag that shows whether this function is implemented. Returning zero because it is not implemented will dominate the minimum calculation for `UNLOCKED` partitions.

adamdossa: @tintinweb As above the default implementation in TM of `getTokensByPartition` now returns the user balance for UNLOCKED to address this.

adamdossa (approved):

satyamakgec:
Note - After merging this into `audit-fixes` branch we have to use a specific contract type variable instead of typecasting in TM

tintinweb:
👍 good to merge. tin/diligence.

PR #680, opened 2019-05-25
Fix the returnPartition function

satyamakgec:


contracts/tokens/SecurityToken.sol
```diff
@@ -527,11 +527,11 @@ contract SecurityToken is ERC20, ReentrancyGuard, SecurityTokenStorage, IERC1594
         // require(balanceOfByPartition(_partition, msg.sender) >= _value);
         // NB - Above condition will be automatically checked using the executeTransfer() function execution.
         // NB - passing `_additionalBalance` value is 0 because accessing the balance before transfer
-        uint256 balanceBeforeTransferLocked = _balanceOfByPartition(_partition, _to, 0);
+        uint256 lockedBalanceBeforeTransfer = _balanceOfByPartition(LOCKED, _to, 0);
```
tintinweb, line 530: argument `_partition` is required to be `UNLOCKED` in `_isValidPartition` but it is not used in `_balanceOfByPartition` which instead hardcodes `LOCKED`. `toPartition` will always be `UNLOCKED` because of a bug in `_returnPartition`. finally an event gets emitted with the unused `_partition` (which needs to be `UNLOCKED` from the function call args.

I am struggling to understand the changes. @satyamakgec I would appreciate if you could elaborate on how this is working.

tintinweb:
Hi @satyamakgec,

1. `_returnPartition`: if line 542 sets `toPartition` to `LOCKED` it is always overridden at line 544 to be `UNLOCKED`.  This is still the case with this PR.
https://github.com/PolymathNetwork/polymath-core/blob/6ff637c4e79a4925fe9636ef0079b1ab4ed4f0e3/contracts/tokens/SecurityToken.sol#L538-L544

From the issue reported in the initial report:
>It's commented in the function that it should return LOCKED under some circumstances.
Function _returnPartition always returns UNLOCKED partition.
Additionally, the comment says that LOCKED should be returned when a change in the locked partition equals value but it actually checks the change in _partition which is usually UNLOCKED

2. as mentioned inline please elaborate a bit on the changes.

thanks!
tin/diligence

contracts/tokens/SecurityToken.sol
satyamakgec: So the Idea is that tokenHolder only allows transacting the `UNLOCKED` token. `_isValidPartition()` is used to validate whether the partition is UNLOCKED or not.  Then function checks the earlier `LOCKED` balance of the `_to` and after transfer, it will check the new `LOCKED` balance of `_to` so `_returnPartition` checks if the difference between both locked balance is equal to the value that is being transferred then it means all value is transferred to `LOCKED` partition. If not then it means some(or full) value transfferd to `UNLOCKED` partition of the receiver(_to).

satyamakgec: And I agree on the 542 line change I should add the else statement.

tintinweb:
👍 tin/diligence

maxsam4 (approved):

PR #683, opened 2019-05-27
Specific contract type used

satyamakgec:


contracts/libraries/TokenLib.sol
```diff
@@ -1,13 +1,13 @@
 pragma solidity ^0.5.0;
 
+import "./KindMath.sol";
```
maxsam4: This may lead to adding back `KindMath` import after the merge.

satyamakgec: I think this will be solved by the other PR.

maxsam4: That pr removes it but this pr ads it back. (it's in green not white). This pr is specifically adding it back.

contracts/libraries/TokenLib.sol
```diff
@@ -231,19 +230,19 @@ library TokenLib {
         address _module,
         uint256 _change,
         bool _increase,
-        address _polyToken,
+        IERC20 _polyToken,
```
tintinweb, line 236: Note: could be `IPoly` to avoid casting on line 242,245 but I see that this is for consistency as polyToken is usually used as IERC20 without ever calling increase/decreaseApproval (which is only part of `IPoly`). Is probably cleaner the way it is right now keeping polyToken an IERC20 and only casting it to IPoly for increaseApproval :)

contracts/tokens/SecurityToken.sol
```diff
@@ -1061,9 +1057,9 @@ contract SecurityToken is ERC20, ReentrancyGuard, SecurityTokenStorage, IERC1594
 
     function updateFromRegistry() public {
         _onlyOwner();
-        moduleRegistry = PolymathRegistry(polymathRegistry).getAddress("ModuleRegistry");
-        securityTokenRegistry = PolymathRegistry(polymathRegistry).getAddress("SecurityTokenRegistry");
-        polyToken = PolymathRegistry(polymathRegistry).getAddress("PolyToken");
+        moduleRegistry = IModuleRegistry(IPolymathRegistry(polymathRegistry).getAddress("ModuleRegistry"));
```
tintinweb: might be unnecessary to cast the registry to `IPolymathRegistry` as it is already declared as `IPolymathRegistry` in `SecurityTokenStorage`.

same in `UpgradeableModuleFactory::76` (not part of this PR) where `polymathRegistry` is a statevar inherited from `ModuleFactory` declared as `IPolymathRegistry` already

tintinweb:
one minor note for what appears to be a double cast - please verify. most other issues (eg. address comparison should be caught by the compiler). good to merge otherwise.

tin/diligence

tintinweb:
👍 tin/diligence

maxsam4 (approved):

maxsam4:
@satyamakgec Please fix the merge conflicts.

PR #684, opened 2019-05-27
Remove KindMath from TokenLib

satyamakgec:


maxsam4 (commented):
Kindmath is not used anywhere any more. Maybe we should delete it from the repo?

tintinweb:
👍 consider removing the file entirely if not used anymore as suggested by maxsam4.

tin/diligence

PR #687, opened 2019-05-27
Remove the IBoot from the contracts

satyamakgec:


maxsam4 (approved):

tintinweb:
✔️good to merge
Note: `VersionUtils.sol` is used in the specific factories and indirectly imported via `ModuleFactory.sol` which might not be very obvious.

PR #693, opened 2019-05-27
Made TM results consistent

maxsam4:
Changed logic so that in the case of no un-archived TMs, all transfers are forbidden (rather than allowed as per current logic).

Fixes Audit Issue 6


tintinweb:
👍 tin/diligence

PR #694, opened 2019-05-27
Audit  3.4 & 3.14 Fix ST upgrades

adamdossa:
3.4 No new ST can be created after implementation upgrade

### DIFF
--- a/contracts/tokens/STFactory.sol
+++ b/contracts/tokens/STFactory.sol
@@ -35,7 +35,7 @@ contract STFactory is ISTFactory, Ownable {
 
     uint256 public latestUpgrade;
 
-    event LogicContractSet(string _version, address _logicContract, bytes _upgradeData);
+    event LogicContractSet(string _version, address _logicContract, bytes _initializationData, bytes _upgradeData);
     event TokenUpgraded(
         address indexed _securityToken,
         uint256 indexed _version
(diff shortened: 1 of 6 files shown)


contracts/tokens/STFactory.sol
```diff
@@ -136,15 +137,18 @@ contract STFactory is ISTFactory, Ownable {
      * @param _logicContract Address of deployed module logic contract referenced from proxy
      * @param _upgradeData Data to be passed in call to upgradeToAndCall when a token upgrades its module
      */
-    function setLogicContract(string calldata _version, address _logicContract, bytes calldata _upgradeData) external onlyOwner {
+    function setLogicContract(string calldata _version, address _logicContract, bytes calldata _initializationData, bytes calldata _upgradeData) external onlyOwner {
         require(keccak256(abi.encodePacked(_version)) != keccak256(abi.encodePacked(logicContracts[latestUpgrade].version)), "Same version");
         require(_logicContract != logicContracts[latestUpgrade].logicContract, "Same version");
         require(_logicContract != address(0), "Invalid address");
+        require(_initializationData.length > 4, "Invalid Initialization");
+        require(_upgradeData.length > 4, "Invalid Upgrade");
```
maxsam4, line 148: Should there be `_upgradeData` for all upgrades? I can see some minor upgrades not needing any upgrade data.

adamdossa: In that case I was thinking we could still specify a no-op function (e.g. that just emits an Upgraded event or does nothing). This seemed "safer" than allowing an empty upgradeData which could be a mistake.

maxsam4 (approved):

tintinweb:
👍 tin/diligence

PR #696, opened 2019-05-28
[3.23] Added break in deleteDelegate

maxsam4:
Minor optimization that reduces the scope of Issue 3.23 but doesn't solve it completely.


satyamakgec (approved):

tintinweb:
👍 tin/diligence

PR #698, opened 2019-05-28
[3.5, 3.7, 3.8] Fix custom modules

adamdossa:
Changes made:

  - made `useModule` and `registerModule` non-reentrant as they make external calls to potentially untrusted contracts
  - store factory owner on initial registration locally and reference this rather than calling back out to `owner` on the module factory. This means that a malicious factory cannot change the value of `owner` dynamically to e.g. allow security tokens to add it (by returning an owner equal the ST owner). The initial `owner` cannot be malicious as `registerModule` requires msg.sender to be the factory owner.
  - fix issue where a malicious factory could return a `type` of 0, causing it to be un-removable
  - modified `getModulesByType` to only return verified modules, so that a malicious unverified module cannot grief this function.
  - added `getAllModulesByType` can be used to return a raw list of all registered modules instead
  - modify `getFactoryDetails` to also return the factory owner

### DIFF
--- a/contracts/ModuleRegistry.sol
+++ b/contracts/ModuleRegistry.sol
@@ -228,8 +228,6 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
         uint256 moduleType = getUintValue(Encoder.getKey("registry", _moduleFactory));
 
         require(moduleType != 0, "Module factory should be registered");
-        // Ordering (lazy evaluation) in below require statement is important. A moduleFactory should not be able
-        // to prevent the MR owner from removing it by throwing on a call to `owner`
         require(
             msg.sender == owner() || msg.sender == getAddressValue(Encoder.getKey("factoryOwner", _moduleFactory)),
             "msg.sender must be the Module Factory owner or registry curator"
@@ -284,8 +282,6 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
         bool isOwner = msg.sender == owner();
         bool isFactory = msg.sender == _moduleFactory;
         bool isFactoryOwner = msg.sender == getAddressValue(Encoder.getKey("factoryOwner", _moduleFactory));
-        // Ordering (lazy evaluation) in below require statement is important. A moduleFactory should not be able
-        // to prevent the MR owner from unverifying it by throwing on a call to `owner`
         require(isOwner || isFactory || isFactoryOwner, "Not authorised");
         require(getUintValue(Encoder.getKey("registry", _moduleFactory)) != uint256(0), "Module factory must be registered");
         set(Encoder.getKey("verified", _moduleFactory), false);
--- a/contracts/ModuleRegistry.sol
+++ b/contracts/ModuleRegistry.sol
@@ -36,6 +36,7 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
     */
 
     bytes32 constant INITIALIZE = 0x9ef7257c3339b099aacf96e55122ee78fb65a36bd2a6c19249882be9c98633bf; //keccak256("initialised")
+    bytes32 constant LOCKED = 0xab99c6d7581cbb37d2e578d3097bfdd3323e05447f1fd7670b6c3a3fb9d9ff79; //keccak256("locked")
     bytes32 constant POLYTOKEN = 0xacf8fbd51bb4b83ba426cdb12f63be74db97c412515797993d2a385542e311d7; //keccak256("polyToken")
     bytes32 constant PAUSED = 0xee35723ac350a69d2a92d3703f17439cbaadf2f093a21ba5bf5f1a53eb2a14d9; //keccak256("paused")
     bytes32 constant OWNER = 0x02016836a56b71f0d02689e69e326f4f4c1b9057164ef592671cf0d37c8040c0; //keccak256("owner")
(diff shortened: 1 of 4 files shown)


contracts/ModuleRegistry.sol
```diff
@@ -129,10 +144,11 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
      * @param _moduleFactory is the address of the relevant module factory
      * @param _isUpgrade whether or not the function is being called as a result of an upgrade
      */
-    function useModule(address _moduleFactory, bool _isUpgrade) external {
-        if (IFeatureRegistry(getAddressValue(FEATURE_REGISTRY)).getFeatureStatus("customModulesAllowed")) {
+    function useModule(address _moduleFactory, bool _isUpgrade) external nonReentrant {
```
maxsam4, line 147: useModule() does not need to be guarded for reentrancy. There is no harm in guarding it either so it's fine anyway.

contracts/ModuleRegistry.sol
```diff
@@ -255,12 +279,14 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
...
     */
-    function unverifyModule(address _moduleFactory) external {
+    function unverifyModule(address _moduleFactory) external nonReentrant {
         // Can be called by the registry owner, the module factory, or the module factory owner
         bool isOwner = msg.sender == owner();
-        bool isFactoryOwner = msg.sender == IOwnable(_moduleFactory).owner();
         bool isFactory = msg.sender == _moduleFactory;
-        require(isOwner || isFactoryOwner || isFactory, "Not authorised");
+        bool isFactoryOwner = msg.sender == getAddressValue(Encoder.getKey("factoryOwner", _moduleFactory));
+        // Ordering (lazy evaluation) in below require statement is important. A moduleFactory should not be able
```
maxsam4: This code comment isn't valid anymore, is it?

contracts/ModuleRegistry.sol
```diff
@@ -323,16 +350,42 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
...
     }
 
     /**
-     * @notice Returns the list of addresses of Module Factory of a particular type
+     * @notice Returns the list of addresses of verified Module Factory of a particular type
      * @param _moduleType Type of Module
      * @return address array that contains the list of addresses of module factory contracts.
      */
     function getModulesByType(uint8 _moduleType) public view returns(address[] memory) {
+        address[] memory _addressList = getArrayAddress(Encoder.getKey("moduleList", uint256(_moduleType)));
```
maxsam4, line 359: This will only return modules with primary type (type[0]) == `_moduleType`. Maybe we should add the factory to all the relevant module type lists in registerModule()?

contracts/ModuleRegistry.sol
```diff
@@ -208,8 +228,10 @@ contract ModuleRegistry is IModuleRegistry, EternalStorage {
         uint256 moduleType = getUintValue(Encoder.getKey("registry", _moduleFactory));
 
         require(moduleType != 0, "Module factory should be registered");
+        // Ordering (lazy evaluation) in below require statement is important. A moduleFactory should not be able
```
maxsam4: This code comment isn't valid anymore, is it?

contracts/ModuleRegistry.sol
maxsam4, line 359: Since we are fetching all modules (verified and unverified) here and looping over them, this function can still be griefed by unverified modules. It will require like 10k unverified modules to make this function cost 8m gas but it can be done.

maxsam4 (commented):
In registerModule() L217, the complete array of modules by type is read from storage and copied into memory. This means once the list grows too big (~15k modules), this function will stop working.

Also, 2 tests are failing.

contracts/ModuleRegistry.sol
adamdossa: `isCompatibleModule` called from `useModule` makes external calls

contracts/ModuleRegistry.sol
adamdossa: True - was from an earlier iteration - removed.

contracts/ModuleRegistry.sol
adamdossa: Yeah - that would probably be better, although we'd also have to modify the `removeModule` logic and the "moduleListIndex" etc. to reflect this, so I'd say this is better in 3.1 than 3.0

contracts/ModuleRegistry.sol
adamdossa: removed

contracts/ModuleRegistry.sol
maxsam4: yes but there is no point in re-entering `useModule()`. I don't see any reason why any (malicious) user might want to renter `useModule()`.

contracts/ModuleRegistry.sol
maxsam4: Agreed. As this function is really only for offchain stuff, I don't mind it returning incomplete data. Perhaps we should add a code comment about this though. I don't want dApp and CLI teams scratching their heads a few months down the line :).

adamdossa: Yep - although the expectation is that this function would only be called off-chain, so could use an even higher gas limit. The bigger concern is in `registerModule` where we read in the whole list which obviously is done on chain so could be griefed. Assuming registering a module costs 100,000 gas and a price of 3GWei it only costs 3.3 ETH to register 10,000 modules, so this is a pretty cheap attack. Having said that, our defence would be to upgrade the module registry and fix (e.g. by removing the bad modules and modifying the code) so probably no-one would be very motivated to do this.

maxsam4: Agreed that this is more of an inconvenience than an Issue. I'd want to fix the `registerModule` issue before allowing custom modules (which may be a while away).

btw the attack will cost ~150 ETH because the cost of registering modules will keep increasing and it will actually require ~15k modules to hit the target. Given the cost, I'd be fine with just adding a code comment about it for now.

tintinweb:
👍 tin/diligence

maxsam4 (approved):

PR #709, opened 2019-06-04
Fix canTransfer spec

adamdossa:


satyamakgec (approved):

contracts/tokens/SecurityToken.sol
```diff
@@ -927,22 +926,21 @@ contract SecurityToken is ERC20, ReentrancyGuard, SecurityTokenStorage, IERC1594
...
      * @return byte Ethereum status code (ESC)
      * @return bytes32 Application specific reason code
      */
-    function canTransferFrom(address _from, address _to, uint256 _value, bytes calldata _data) external view returns (bool success, byte reasonCode, bytes32 appCode) {
-        (success, reasonCode, appCode) = _canTransfer(_from, _to, _value, _data);
-        if (success && _value > allowance(_from, msg.sender)) {
-            return (false, 0x53, bytes32(0));
+    function canTransferFrom(address _from, address _to, uint256 _value, bytes calldata _data) external view returns (byte reasonCode, bytes32 appCode) {
+        (reasonCode, appCode) = _canTransfer(_from, _to, _value, _data);
+        if ((reasonCode & 0x0F) == 0x01 && _value > allowance(_from, msg.sender)) {
```
tintinweb: it should be obvious that this indicates success. could make sense to add either a comment or use the `isSuccess` method you mentioned with the PR details:

> From the ESC-1066 reference implementation we have:
 ```
/**
 * @dev Check if a status code represents success (ie: 0x*1)
 *
 * @param status Binary ERC-1066 status code
 * @return successful A boolean representing if the status code
 *         represents success
 */
function isSuccess(byte status) public pure returns (bool successful) {
    return reasonOf(status) == 0x01;
}
```

also note that https://github.com/PolymathNetwork/polymath-core/pull/690 introduces named status codes and might conflict with this PR (yours is merging to dev-3.0.0; #690 is merged to `audit-fixes` which is pending a merge to dev-3.0.0 (#705))

another option could to check against  the fixed value `status.TransferSuccess` (`0x51`) instead of the least significant nibble being 0x1 (unless 0x11, 0x21, 0x31 can be a valid success esc as noted in your quote)

File: test/j_manual_approval_transfer_manager.js [hunk 2]
```diff
@@ -397,17 +397,19 @@ contract("ManualApprovalTransferManager", accounts => {
                 }
             );
             console.log(JSON.stringify(verified[0]));
-            assert.equal(verified[0], true);
+            assert.equal(verified[0], 0x51);
```
tintinweb: nb (readability): may use enum as in https://github.com/PolymathNetwork/polymath-core/pull/690/files

test/j_manual_approval_transfer_manager.js
```diff
@@ -397,17 +397,19 @@ contract("ManualApprovalTransferManager", accounts => {
...
 
             verified = await I_SecurityToken.canTransfer.call(account_investor4, web3.utils.toWei("4", "ether"), "0x0", {
                 from: account_investor1
             });
-            assert.equal(verified[0], false);
+            console.log(JSON.stringify(verified[0]));
+//            assert.equal(verified[0], false);
```
tintinweb: test w/o validation. this and the one below should be checked against an expected error code.

contracts/tokens/SecurityToken.sol
adamdossa: Modified to use a helper `isSuccess` function to check the LS nibble.

test/j_manual_approval_transfer_manager.js
adamdossa: fixed

tintinweb:
👍

PR #719, opened 2019-06-17
Add isVolRestricted in the verifyTransfer

satyamakgec:


maxsam4:
The solidity code looks right but missing test cases. I have one question though. Do we really need `isVolRestricted` ? It seems kind of useless to me. Why will an issuer add individual restriction details for an investor and them mark them as not `isVolRestricted`.

It will make sense if it worked as the exemption list and exempted the investor from global default restriction as well.

ps We'll have to get these changes approved from CD.

cc @adamdossa @pabloruiz55

adamdossa:
@satyamakgec closing this as we're no longer doing this.

PR #731, opened 2019-06-22
ISTR script

maxsam4:
This script makes sure that ISTR is in sync with STR + STRGetter.

// The CI will fail right now because ISTR is out of sync.


satyamakgec (approved):
You can use local truffle for compilation instead of the global truffle. This will help like me who have a 4.14 global truffle version.  But anyway it will run on the remote that will have required truffle version.

tintinweb:
👍 nice way to keep things in sync.

PR #735, opened 2019-07-03
Added Upper bounds to 2.x modules

maxsam4:
Rather than using a wildcard 0.0.0, this PR changes the upper bound of 2.x modules to 2.9.9.


satyamakgec:
@maxsam4 could you please change the ETHUSDOracle address as well in the deploy script. 
This is the new Kovan ETH/USD Oracle - https://kovan.etherscan.io/address/0x14542627196c7dab26eb11ffd8a407ffc476de76

satyamakgec (approved):

PR #736, opened 2019-07-03
Fixed CTM verifyTransfer bug

maxsam4:
Fixed a bug found by Victor

>I found something that seems to be a bug.
I have added a CTM with `maxHolders = 2`
My token has 2 holders:
A -> 800 tokens
B -> 200 tokens
If I call to `canTransfer` on ST20 to check if A can transfer 100 tokens to C, it returns `TransferSuccess`.
Same if I call to `verifyTransfer` on my CTM. It doesn’t return `INVALID`.
But if I actually try to do `transfer` it is reverted.

### DIFF
--- a/contracts/modules/TransferManager/CTM/CountTransferManager.sol
+++ b/contracts/modules/TransferManager/CTM/CountTransferManager.sol
@@ -33,7 +33,7 @@ contract CountTransferManager is CountTransferManagerStorage, TransferManager {
         external
         returns(Result)
     {
-        (Result success, ) = _verifyTransfer(_from, _to, _amount);
+        (Result success, ) = _verifyTransfer(_from, _to, _amount, securityToken.holderCount());
         return success;
     }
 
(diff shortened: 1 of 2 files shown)


adamdossa (approved):

satyamakgec (approved):

contracts/modules/TransferManager/CTM/CountTransferManager.sol
```diff
@@ -53,20 +53,33 @@ contract CountTransferManager is CountTransferManagerStorage, TransferManager {
...
-        return _verifyTransfer(_from, _to, _amount);
+        uint256 holderCount = securityToken.holderCount();
+        if (_amount != 0 && _from != _to) {
+            // Check whether receiver is a new token holder
+            if (_to != address(0) && securityToken.balanceOf(_to) == 0) {
+                holderCount++;
+            }
+            // Check whether sender is moving all of their tokens
+            if (_amount == securityToken.balanceOf(_from)) {
+                holderCount--;
```
tintinweb, line 68: guard `holderCount` from over/underflows as it can be zero. See existing implementation in `TokenLib.adjustInvestorCount`.

tintinweb:
Something that could be mentioned in a comment because otherwise the checks might look inconsistent at first glance (`executeTransfer` performing a different check than `verifyTransfer`):

* `module.verifyTransfer` is called by `SecToken.canTransfer` and does not receive the updated `holderCount` therefore `verifyTransfer` has to manually account for pot. tokenholder changes (by mimicking `TokenLib.adjustInvestorCount`).
* `module.executeTransfer` is called by `SecToken.transfer|issue|others` and receives an updated `holderCount` as `sectoken` calls `TokenLib.adjustInvestorCount` before `executeTransfer`.

contracts/modules/TransferManager/CTM/CountTransferManager.sol
maxsam4: @tintinweb Since `holderCount` can only be incremented/decremented by one at a time, it can not be overflowed.

`holderCount` cannot be underflowed either since `holderCount` must be greater than 1 if `_amount != 0 && _from != _to` (checked on L57).

Earlier, we used SafeMath everywhere (including `TokenLib.adjustInvestorCount`) but lately, I have using SafeMath only where it is needed (for optimization). 

Do you have a scenario in mind where holderCount can be underflowed or overflowed?

tintinweb: thanks @maxsam4, agree the upper bound is unrealistic (especially when using ctm :)) and `holderCount==0` with `amount !=0` is mutually exclusive 👍

PR #737, opened 2019-07-03
Move some variables / functions to internal.

adamdossa:
Move some variables / functions to internal


maxsam4 (approved):
Changes look good to me.

I am not sure if this will solve the problem @monitz87 was facing or not though.

satyamakgec (approved):

PR #738, opened 2019-07-03
TakeFee was removed.

onigiri-x:
### What is the new behavior?
It seems the changelog doesn't reflect takeFee was removed from Module.sol


monitz87 (approved):
Looks good

adamdossa (approved):

PR #757, opened 2019-07-18
Allow pre-minting in the STOs

satyamakgec:


contracts/modules/STO/STO.sol
```diff
@@ -55,4 +55,20 @@ contract STO is ISTO, STOStorage, Module {
     function _getKey(bytes32 _key1, address _key2) internal pure returns(bytes32) {
         return bytes32(keccak256(abi.encodePacked(_key1, _key2)));
     }
+
+    function _allowPreMinting(uint256 _tokenAmount) internal {
```
adamdossa, line 59: Should we allow users to turn this off after turning it on? If so, I guess we'd either have to move the already issued to the treasury wallet, or burn them (I'd favour the latter).

contracts/modules/STO/Capped/CappedSTO.sol
```diff
@@ -258,4 +273,26 @@ contract CappedSTO is CappedSTOStorage, STO, ReentrancyGuard {
         polyToken.transferFrom(_beneficiary, _to, _fundsAmount);
     }
 
+    /**
+     * @notice Finalizes the STO and mint remaining tokens to treasury address
+     * @notice Treasury wallet address must be whitelisted to successfully finalize
+     */
+    function finalize() external withPerm(ADMIN){
+        require(!isFinalized, "STO is finalized");
+        isFinalized = true;
```
adamdossa, line 293: Should we add `require(!isFinalized)` to `_preValidatePurchase`? I don't think you can buy tokens once the STO has been finalized.

contracts/modules/STO/Capped/CappedSTO.sol
```diff
@@ -76,6 +81,13 @@ contract CappedSTO is CappedSTOStorage, STO, ReentrancyGuard {
         return this.configure.selector;
     }
 
+    /**
+     * @notice This function will allow STO to pre-mint all tokens those will be distributed in sale
+     */
+    function allowPreMinting() external withPerm(ADMIN) {
+        _allowPreMinting(cap);
```
adamdossa, line 90: Should this flag be added to `getSTODetails`?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -231,6 +244,12 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
             require(_ratePerTierDiscountPoly[i] <= _ratePerTier[i], "Invalid discount");
             tiers.push(Tier(_ratePerTier[i], _ratePerTierDiscountPoly[i], _tokensPerTierTotal[i], _tokensPerTierDiscountPoly[i], 0, 0));
         }
+        if (preMintAllowed) {
+            uint256 oldCap = securityToken.balanceOf(address(this));
+            uint256 newCap = _getTotalTokensCap();
+            if (oldCap < newCap)
+                securityToken.issue(address(this), (newCap - oldCap), "");
```
adamdossa, line 254: Don't we need to burn (redeem) tokens if `oldCap > newCap` as well?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -278,20 +297,28 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
...
             if (remainingTokens > 0) {
-                tiers[i].mintedTotal = tiers[i].tokenTotal;
+                tiers[i].totalTokensSoldInTier = tiers[i].tokenTotal;
             }
         }
-        address walletAddress = (treasuryWallet == address(0) ? IDataStore(getDataStore()).getAddress(TREASURY) : treasuryWallet);
+        address walletAddress = getTreasuryWallet();
         require(walletAddress != address(0), "Invalid address");
+        if (preMintAllowed)
+            tempReturned = securityToken.balanceOf(address(this));
```
adamdossa: In practice `tempReturned == securityToken.balanceOf(address(this))` before this assignment right? (maybe apart from dust rounding issues)?

contracts/modules/STO/Capped/CappedSTO.sol
```diff
@@ -42,14 +44,16 @@ contract CappedSTO is CappedSTOStorage, STO, ReentrancyGuard {
...
      */
     function configure(
         uint256 _startTime,
         uint256 _endTime,
         uint256 _cap,
         uint256 _rate,
         FundRaiseType[] memory _fundRaiseTypes,
-        address payable _fundsReceiver
+        address payable _fundsReceiver,
+        address _treasuryWallet
```
adamdossa, line 58: Should we add the preMint boolean here as well so it can be set on initialisation? (same for USDTieredSTO)

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -278,20 +297,28 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
...
         uint256 granularity = securityToken.granularity();
         tempReturned = tempReturned.div(granularity);
         tempReturned = tempReturned.mul(granularity);
-        securityToken.issue(walletAddress, tempReturned, "");
-        emit ReserveTokenMint(msg.sender, walletAddress, tempReturned, currentTier);
+        if (tempReturned != uint256(0)) {
+            if (preMintAllowed) {
+                securityToken.transfer(walletAddress, tempReturned);
```
adamdossa, line 325: Maybe we should have an event emitted here as well - i.e. `ReserveTokenTransfer` - this is already kind of captured with the token transfer itself, but maybe good to be explicit that it has been finalized.

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -188,6 +190,22 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
...
     }
 
+    function _isValidModule(address _holder) internal view returns(bool) {
+        uint8[] memory types;
+        bool isArchived;
+        bytes32 name;
+        // Use getModule() function to avoid the malicious module change its factory type after adding
+        // into the securityToken 
+        (name,,,isArchived,types,)= securityToken.getModule(_holder);
+        if (name != bytes32(0) && !isArchived) {
```
adamdossa: Why do we care about `name != bytes32(0)`?

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -188,6 +190,22 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
         return (Result.NA, bytes32(0));
     }
 
+    function _isValidModule(address _holder) internal view returns(bool) {
```
adamdossa: can we change name to _isWhitelistModule

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
```diff
@@ -171,7 +171,9 @@ contract GeneralTransferManager is GeneralTransferManagerStorage, TransferManage
                 txReq = transferRequirements[uint8(TransferType.GENERAL)];
             }
 
-            (canSendAfter, fromExpiry, canReceiveAfter, toExpiry) = _getValuesForTransfer(_from, _to);
+            !_isValidModule(_from) ? (canSendAfter, fromExpiry) = _getValuesForFrom(_from) : (canSendAfter, fromExpiry) = (uint64(now - 1), uint64(now + 1));
```
adamdossa: Doesn't really matter, but maybe use (0, and maxUint64) instead here?

contracts/modules/STO/STO.sol
satyamakgec: I don't think this is necessary, It will add complexity in the code (+ code size) nothing else. The issuer has the clear choice of where to go at the time of launch of the STO either pre-mint option or mint on buying option.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
satyamakgec: Yes, this is a possibility It is an open question about how the issuer wants? Right now all extra tokens are extracted at the time of finalization of the STO. **[Product Question]**

contracts/modules/STO/Capped/CappedSTO.sol
satyamakgec: Fixed

contracts/modules/STO/Capped/CappedSTO.sol
satyamakgec: Fixed

contracts/modules/STO/USDTiered/USDTieredSTO.sol
satyamakgec: I don't think so, tempReturned will be 0 before the assignment if the (preMintAllowed == true). and after that, I am tackling the rounding issues by making it granular. Or you are suggesting something else?

contracts/modules/STO/Capped/CappedSTO.sol
satyamakgec: I don't think so, because if we add the `preMint` boolean here then we will be spending more gas when issuer chooses `mint on buying`. Anyway, the issuer has to call the `allowPreMint()` function to facilitate the pre-mint functionality so it doesn't make too much sense.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
satyamakgec: Fixed

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
satyamakgec: Fixed

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
satyamakgec: The idea was to avoid going into the loop when the given address is not the module.

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
satyamakgec: How 0 will work for `canSendAfter & canRecieveAfter`? but yeah maxUint64 will work for expiry

contracts/modules/STO/Capped/CappedSTO.sol
adamdossa: I was assuming that if `preMint` was set to true, then `_allowPreMint` would be called on initialisation automatically (i.e. it becomes a one-step process, rather than two-step process). You're right that if `preMint` was false it would be slightly more expensive than just having this as a default. Have asked @Mawadda-Poly this question over slack.

contracts/modules/STO/STO.sol
adamdossa: Checking w/ Product on this.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
adamdossa: Why would it be zero? I think `tiers[i].tokenTotal` and `tiers[i].totalTokensSoldInTier` are still updated during the STO if `preMintAllowed == true`?

adamdossa: I guess we need this if the cap was decreased before the STO starts (in which case the balance may not match the difference between the cap(s) and the number of tokens sold)?

contracts/modules/TransferManager/GTM/GeneralTransferManager.sol
adamdossa: Probably better to check `module` returned value, rather than `name` then I think.

adamdossa (approved):

PR #777, opened 2019-07-31
Enhance "branches" test coverage of USDTieredSTO

ross-rosario:
This PR brings a few more tests to increase the coverage as mentioned above. 

Coverage is still not 100% though. Some uncovered lines are just not worth testing because they're too obvious, or are just unreachable. For sake of clarity, I put a `@dev` comment on those unreachable branches in the smart contract.

I had captured the coverage report for said contract and attached it below. Once CI coverage is generated, you can compare both pre and post coverages.

[USDTieredSTO.sol.html.zip](https://github.com/PolymathNetwork/polymath-core/files/3451226/USDTieredSTO.sol.html.zip)

### DIFF
--- a/contracts/modules/STO/USDTiered/USDTieredSTO.sol
+++ b/contracts/modules/STO/USDTiered/USDTieredSTO.sol
@@ -56,12 +56,16 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
     ///////////////
 
     modifier validETH() {
+        // @dev "Else" branch is unreachable. In case of 0x0 custom oracle, _getOracle() will return the
+        // default one from polymathRegistry.
         require(_getOracle(bytes32("ETH"), bytes32("USD")) != address(0), "Invalid Oracle");
         require(fundRaiseTypes[uint8(FundRaiseType.ETH)], "ETH not allowed");
         _;
     }
 
     modifier validPOLY() {
+         // @dev "Else" branch is unreachable. In case of 0x0 custom oracle, _getOracle() will return the
+        // default one from polymathRegistry.
         require(_getOracle(bytes32("POLY"), bytes32("USD")) != address(0), "Invalid Oracle");
         require(fundRaiseTypes[uint8(FundRaiseType.POLY)], "POLY not allowed");
         _;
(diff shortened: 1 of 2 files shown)


ross-rosario:
I just realized that CI coverage artifacts might not be easy to access. So I attached the locally generated report to easily compare it with the original report.

[USDTieredSTO.sol.html.zip](https://github.com/PolymathNetwork/polymath-core/files/3451335/USDTieredSTO.sol.html.zip)

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -286,6 +291,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
             }
         }
         address walletAddress = (treasuryWallet == address(0) ? IDataStore(getDataStore()).getAddress(TREASURY) : treasuryWallet);
+        // @dev "else" branch is unreachable.
```
maxsam4: The treasury address can be set to `0x0` to cover this branch. Similar techniques can be applied to some other unreachable branched. That being said, Unless we are aiming for 100% coverage, I am fine with these branches not being covered.

maxsam4 (approved):

contracts/modules/STO/USDTiered/USDTieredSTO.sol
ross-rosario: To clarify, my point is, there's no way to have the following `require` statement revert because `walletAddress` cannot be `0x0`, given the conditions above.

maxsam4: To clarify, my point is, there's a still a way to trigger the require statement :)

The above statement says that 
if treasury wallet is not zero, set wallet address to treasury wallet
but if
treasury wallet  is zero then fetch proper address from the datastore

Now, what if, we set treasury wallet = 0 so that second statement is triggered and we also set the address in datastore to 0? That way the contract will query the datastore but the datastore will return 0x0 so the wallet address will in turn be 0x0.

ross-rosario: Yeah, but DataStore has its own checks that prevent setting treasury to 0x0 right?

maxsam4: Nope, it's a dumb contract :)

satyamakgec: There is no explicit check in the dataStore to avoid 0x0 as an address to store. But you can't do it by using change `changeTreasuryWallet()` function of the ST because we have 0x0 check while you have to manage to call `setAddress()` function of the dataStore directly to set globally treasury wallet to 0x0.

satyamakgec (approved):

PR #783, opened 2019-08-06
Dependency upgrade

maxsam4:


satyamakgec (approved):

ross-rosario:
Thanks for removing the coverage safeguards!

@maxsam4 I see that both `yarn.lock` and `package-lock.json` have been updated. Shall we just remove one in favor of the other? I'm personally indifferent (though `npm` might make more sense since it's already shipped with node.)

**Edit**: `package-lock.json` is being brought back to hopefully, tackle the stagnant vulnerabilities report of github.

maxsam4:
> Thanks for removing the coverage safeguards!
> 
> @maxsam4 I see that both `yarn.lock` and `package-lock.json` have been updated. Shall we just remove one in favor of the other? I'm personally indifferent (though `npm` might make more sense since it's already shipped with node.)

The only reason I added both is to update the cached file in GitHub security. We can delete it later. More details on slack :)

PR #794, opened 2019-08-20
Reinventing voting

satyamakgec:


contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,1004 @@
...
+            _noOfChoices,
+            _checkpointId
+        );         
+        _addExemptedAddresses(_exemptedAddresses, ballots.length - 1);
+    } 
+
+    function _addExemptedAddresses(address[] memory _exemptedAddresses, uint256 _ballotId) internal {
+        for (uint256 i = 0; i < _exemptedAddresses.length; i++) {
+            Ballot storage ballot = ballots[_ballotId];
+            if (_exemptedAddresses[i] != address(0))
```
maxsam4, line 297: do we need this check?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,1004 @@
...
+        _checkValidStage(_ballotId, Stage.REVEAL);
+        Ballot storage ballot = ballots[_ballotId];
+        // validate the storage values 
+        require(!ballot.isCancelled, "Cancelled ballot");
+        require(ballot.voteDetails[msg.sender].secretVote != bytes32(0), "Secret vote not available");
+        uint256 choiceCount = 0;
+        uint256 i;
+        for (i = 0; i < ballot.totalProposals; i++) {
+            uint256 _noOfChoice = ballot.proposals[i].noOfChoices;
+            _noOfChoice = _noOfChoice == 0 ? 3 : _noOfChoice;
```
maxsam4, line 465: why is 3 the default?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: Yeah, No hard need just don't want unnecessary storage change.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: 3 is used when your proposal is NAY/YAY type so when you will create the proposal with NAY/YAY type you will just pass no. of choices = 0 so the contract will automatically assign 3 choices to it.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,1003 @@
...
+    }
+
+    /**
+     * Use to check whether the voter is allowed to vote or not
+     * @param _ballotId The index of the target ballot
+     * @param _voter Address of the voter
+     * @return bool
+     */
+    function isVoterAllowed(uint256 _ballotId, address _voter) public view returns(bool) {
+        bool allowed = (ballots[_ballotId].exemptedVoters[_voter] || (defaultExemptIndex[_voter] != 0));
```
adamdossa: Variable should probably be called `notAllowed`

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
adamdossa: Maybe better to store 3 as default (rather than 0) when setting `noOfChoice` rather than here?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -0,0 +1,1003 @@
...
+            for (i = 0; i < ballot.totalProposals; i++) {
+                uint256 _noOfChoices = ballot.proposals[i].noOfChoices;
+                _noOfChoices = _noOfChoices == 0 ? 3 : _noOfChoices;
+                noOfChoicesInProposal[i] = _noOfChoices;
+                uint256 nextProposalChoiceLen = count + _noOfChoices;
+                for (j = 0; j < ballot.totalVoters; j++) {
+                    uint256[] storage choiceWeight = ballot.voteDetails[voters[j]].voteOptions[i];
+                    uint256 l = 0;
+                    for (k = count; k < nextProposalChoiceLen; k++) {
+                        choicesWeighting[k] = choicesWeighting[k].add(choiceWeight[l]);
```
adamdossa, line 885: I think this means that it is unlikely we would be able to retrieve the ballot results on-chain (assuming a large number of voters) - is it possible to keep some totals as people vote so that we could retrieve the results on-chain (e.g. so that a vote could be used to trigger some other on-chain action)?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: If I store 3 as default then there is no way to dApp developers to know that proposal is NAY/YAY type. if there are 3 proposals with choices 3, 0, 3 but the smart contract will return 3, 3, 3 then during parsing the event data `_choices` it will be very hard to know which proposal is NAY/YAY type whether first or second or third. 
One thing I can do is create a constant variable with 3 as value and use that everywhere for better understanding.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: Yes, you're right it is not suitable to call this function on-chain. I don't think any easy and clean way to do it beside this - 
Rather keeping the weight of the choices per investor we will in parallel calculating the result (sum of weight for a choice of a proposal) of the vote and store it so when we will query the `getBallotResults()` function we will just return it. The downside of this mechanism is the increased gas cost of `revealVote()` function and It may change the functionality of many getters that are placed in the module.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -530,22 +526,53 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
...
+     */
+    function changeDefaultExemptedVotersListMulti(address[] memory _voters, bool[] memory _exempts) public {
+        require(!_isAnyBallotRunning());
+        super.changeDefaultExemptedVotersListMulti(_voters, _exempts);
+    }
+
+    function _isAnyBallotRunning() internal view returns (bool isAnyBallotActive) {
+        uint256 length = ballots.length;
+        isAnyBallotActive = false;
+        for (uint256 i = 0; i < length; i++) {
```
adamdossa: @satyamakgec this is a bit scary with an unbounded loop. Is it possible to keep a counter (in global state) of active ballots instead of looping?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -530,22 +526,53 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
...
+            if (getCurrentBallotStage(i) != Stage.PREP || getCurrentBallotStage(i) != Stage.RESOLVED) {
```
adamdossa: Would it be clearer to use && over {COMMIT, REVEAL} instead?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: There is no other way to do it until unless we are comfortable to allow issuer manually end the ballot. IMO this will not affect too much first it is a least callable function.

I reversed the loop as the biggest ballotId has more chances to be in the active stage, this will not completely remove the threat but will lower it.

adamdossa (approved):

PR #811, opened 2019-09-10
Skipping treasury Wallet

satyamakgec:


adamdossa (approved):
Looks good to me. One question - we currently have:
1. Modules which implement ITreasuryModule (WHITELISTTREASURY) and can have their `getTreasuryWallet` automatically whitelisted.
2. Modules which are of type WHITELISTMODULE where the module itself is automatically whitelisted.

Do any modules use both these approaches? The 2nd could be collapsed into the 1st by having WHITELISTMODULE's return `address(this)` as their `getTreasuryWallet`.

If we expect both cases to be needed in a single module, then the current logic makes sense though, so this is just a comment.

PR #816, opened 2019-09-19
Issuance module

satyamakgec:


adamdossa (approved):
Do we really need the `IssuanceStorage` contract given it's empty? Does this make upgrades easier in the future?

satyamakgec:
No hard requirement, It is only to follow the same architecture of modules.

PR #823, opened 2019-09-27
Coverage: more coverage for APLCR voting module

ross-rosario:
Besides the additional tests, I added a few questions that are annotated with `@FIXME` comment.

### DIFF
--- a/contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
+++ b/contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
@@ -296,10 +296,6 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
             Ballot storage ballot = ballots[_ballotId];
             if (_exemptedAddresses[i] != address(0))
                 ballot.exemptedVoters[_exemptedAddresses[i]] = true;
-            // @FIXME uncomment this once we decrease contract size.
-            // else {
-            //     delete _exemptedAddresses[i];
-            // }
         }
         emit VotersExempted(_ballotId, _exemptedAddresses);
     }
@@ -572,11 +568,11 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
         uint256 length = ballots.length;
         isAnyBallotActive = false;
         if (length != 0) {
-            uint256 count = length -1;
-            for (uint256 i = count; i >= 0 && i < count; i--) {
+            uint256 count = length - 1;
+            // "i <= count" is to prevert underflow.
+            for (uint256 i = count; i >= 0 && i <= count; i--) {
                 Stage currentStage = getCurrentBallotStage(i);
-                // @FIXME shouldn't we be ORing commit and reveal stages, instead?
-                if (currentStage == Stage.COMMIT && currentStage == Stage.REVEAL && !ballots[i].isCancelled) {
+                if (!ballots[i].isCancelled && (currentStage == Stage.COMMIT || currentStage == Stage.REVEAL)) {
                     isAnyBallotActive = true;
                     break;
                 }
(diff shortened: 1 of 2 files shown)


contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -602,6 +608,9 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
      * @return address[] list of invesotrs who are remain to vote
      */
     function getPendingInvestorToVote(uint256 _ballotId) external view returns(address[] memory pendingInvestors) {
+        // @TODO for clarity, shall we revert instead of returning an empty array?
+        // if so, can we change it into a modifier?
```
satyamakgec: For the ease of handling on the dApp side, we try to return 0 values instead of revert error @VictorVicente @FahdW what do you prefer?

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -294,8 +294,12 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
     function _addExemptedAddresses(address[] memory _exemptedAddresses, uint256 _ballotId) internal {
         for (uint256 i = 0; i < _exemptedAddresses.length; i++) {
             Ballot storage ballot = ballots[_ballotId];
-            if (_exemptedAddresses[i] != address(0)) 
+            if (_exemptedAddresses[i] != address(0))
                 ballot.exemptedVoters[_exemptedAddresses[i]] = true;
+            // @FIXME uncomment this once we decrease contract size.
+            // else {
+            //     delete _exemptedAddresses[i];
```
satyamakgec: delete makes that particular index value will be 0 and in our case, it is automatically 0 so it doesn't make sense too much

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -570,6 +575,7 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
             uint256 count = length -1;
             for (uint256 i = count; i >= 0 && i < count; i--) {
                 Stage currentStage = getCurrentBallotStage(i);
+                // @FIXME shouldn't we be ORing commit and reveal stages, instead?
```
satyamakgec: sorry, I didn't get what you are trying to say here

satyamakgec (changes requested):

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
ross-rosario: The statement below is `true` if `currentStage == Stage.COMMIT` and `currentStage == Stage.REVEAL`, so it requires the stage to be both commit and reveal at the same time. That's why I wonder if the `&&` should be changed to `||`.

satyamakgec: Yeah it should be `||`, how come I left this and it passes the test as well. Thanks @remon-nashid for pointing this out.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
ross-rosario: Makes sense. Let's stick to returning empty arrays.

ross-rosario:
@satyamakgec this is ready for another review.

satyamakgec (approved):

PR #831, opened 2019-10-14
Usage cost implementation

satyamakgec:


contracts/mocks/Dummy/DummySTOFactory.sol
```diff
@@ -11,18 +11,20 @@ contract DummySTOFactory is UpgradableModuleFactory {
...
     constructor (
         uint256 _setupCost,
+        uint256 _usageCost,
         address _logicContract,
         address _polymathRegistry,
         bool _isCostInPoly
     )
         public
-        UpgradableModuleFactory("3.0.0", _setupCost, _logicContract, _polymathRegistry, _isCostInPoly)
+        UpgradableModuleFactory("3.0.0", _setupCost, _usageCost, _logicContract, _polymathRegistry, _isCostInPoly)
```
pabloruiz55: Should say "3.1.0"

contracts/modules/Module.sol
```diff
@@ -81,12 +85,25 @@ contract Module is IModule, ModuleStorage, Pausable {
...
     function reclaimETH() external {
         _onlySecurityTokenOwner();
         msg.sender.transfer(address(this).balance);
     }
+
+    /**
+     * @notice Deduct the usage cost
+     */
+    function _deductUsageFee() internal {
+        uint256 _usageCost = IModuleFactory(factory).usageCostInPoly();
```
maxsam4, line 101: This means that the factory owner can arbitrarily change usage cost. Front running attacks are also possible where the issuer initiates the transaction thinking that fee is X but module owner front runs that tx to increase the usage fee to Y.

maxsam4: The issuer can reduce the risk by properly using the budgeting in the security token but I think we can come up with a better solution.

contracts/modules/ModuleFactory.sol
```diff
@@ -28,6 +31,9 @@ contract ModuleFactory is IModuleFactory, Ownable {
 
     bool public isCostInPoly;
     uint256 public setupCost;
+    uint256 public usageCost;
+    uint256 public lastTimeUsageCostChange;
+    uint256 internal constant LOCKPERIOD = 24 hours;
```
maxsam4: I am not sure how/why LOCKPERIOD is useful in its current form.

Perhaps we should do something like
1) user can change fee at any time.
2) new fee will become active only after lockperiod.

contracts/modules/Module.sol
satyamakgec: lockup period is used to avoid front running issues. Factory owner can change the value of usage cost in 24 hours only because of this front running chances are very low.

maxsam4: In its current implementation, the lock period is failing to do that. It's just limiting the module owner to do one front-run attack per 24 hours. Ideally, Even one attack shouldn't be possible.

If we implement something like what I mentioned in the other review comment, both this and that review comment will be solved.

contracts/modules/Module.sol
```diff
@@ -81,12 +85,25 @@ contract Module is IModule, ModuleStorage, Pausable {
...
+        if (_usageCost > 0) {
+            address factoryOwner = IModuleFactory(factory).owner();
+            require(factoryOwner != address(0), "Invalid owner");
```
adamdossa: I guess this means we can break modules (that have usage fees) by setting the factoryOwner to 0x0? I wonder whether we should just add this constraint to `if (_usageCost > 0)` instead so that if someone sets the factory owner to 0x0 they lose fees, but don't break the module?

contracts/modules/ModuleFactory.sol
adamdossa: Agreed - should be a two step propose:
  - factory owner proposes a usage fee change
  - factory owner can commit / finalise the change 24 hours after it has been proposed

Exception is when the module is first created of course.

adamdossa: NB - if we did this for setup & usage costs, we could get rid of the logic in ST where the user has to pass an expected setup amount as well - not sure if this is worth fiddling with at this point.

contracts/modules/Experimental/TransferManager/SignedTransferManagerFactory.sol
```diff
@@ -10,13 +10,18 @@ contract SignedTransferManagerFactory is ModuleFactory {
...
+     * @param _isCostInPoly true = cost in Poly, false = USD
      */
     constructor(
         uint256 _setupCost,
+        uint256 _usageCost,
         address _polymathRegistry,
         bool _isCostInPoly
     )
-        public ModuleFactory(_setupCost, _polymathRegistry, _isCostInPoly)
+        public ModuleFactory(_setupCost, _usageCost, _polymathRegistry, _isCostInPoly)
```
adamdossa, line 24: General comment is that with this design we can only have one usage fee per module - so any actions in that module must cost the same amount of POLY. If this is OK w/ product that's fine, but not 100% sure if that's the case or not.

contracts/interfaces/IModuleFactory.sol
```diff
@@ -115,4 +133,11 @@ interface IModuleFactory {
      */
     function changeTitle(string calldata _title) external;
 
+    function polymathRegistry() external returns (address polymathRegistryAddress);
```
adamdossa: Where is this used? Quick search doesn't show anything.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
```diff
@@ -485,6 +486,7 @@ contract AdvancedPLCRVotingCheckpoint is AdvancedPLCRVotingCheckpointStorage, Vo
             }
             choiceCount = temp;
         }
+        require(totalChoiceWeight <= weight);
```
adamdossa, line 489: Is this change due to moving some of the voting stuff to a library, or a separate bug you found?

contracts/modules/Experimental/TransferManager/SignedTransferManagerFactory.sol
satyamakgec: I do have a similar doubt in my mind, but if you see all our modules they are pretty focussed on one major activity and revolves around that. Still, it is a good question to ask to the business team.

contracts/modules/Checkpoint/Voting/APLCR/AdvancedPLCRVotingCheckpoint.sol
satyamakgec: Yeah it is related to the separate bug I found

contracts/interfaces/IModuleFactory.sol
satyamakgec: I used this in the earlier implementation but afterwards requirements get change. I keep this because it doesn't hurt anything and make it compliant with ModuleFactory contract.

contracts/modules/Module.sol
satyamakgec: That makes sense

contracts/modules/Module.sol
satyamakgec: hmm, Both cases are probabilistic but anyway I don't have any problem to implement.

contracts/modules/ModuleFactory.sol
```diff
@@ -94,10 +95,17 @@ contract ModuleFactory is IModuleFactory, Ownable {
...
+        emit ChangeUsageCost(usageFee, _usageCost);
+        tempUsageFee = usageFee;
+        usageFee = _usageCost;
+    }
+
+    function usageCost() public view returns(uint256 usdUsageCost) {
+        if (now > lastTimeUsageCostChange.add(COLDPERIOD))
+            return usageFee;
+        else
+            return tempUsageFee;
```
maxsam4: This can be bypassed by changing the fee twice. We should modify the change usage cost function to only update the tempUsageFee if `now > lastTimeUsageCostChange.add(COLDPERIOD)`.

adamdossa: Yeah - I tend to think of this as needing two operations from the factory owner:
1. set proposed new usage fee (emit UsageFeeProposed(proposedFee, currentFee))
2. confirm proposed usage fee (can only be called more than 24 hours after 1 above) (emit ChangeUsageCost(oldFee, newFee))

This means that if 2. never happens, the fee is never changed. I see this as something that would happen rarely, so not an issue that there needs to be two calls.

contracts/modules/ModuleFactory.sol
```diff
@@ -88,24 +89,26 @@ contract ModuleFactory is IModuleFactory, Ownable {
 
     /**
      * @notice Used to change the usage cost
-     * @param _usageCost new usage cost
      */
-    function changeUsageCost(uint256 _usageCost) public onlyOwner {
-        _changeUsageCost(_usageCost);
+    function changeUsageCost() public onlyOwner {
+        require(now > usageCostProposedAt.add(COLDPERIOD), "Proposal is in unmatured state");
+        require(usageCost != proposedUsageCost);
```
maxsam4: This check seems unnecessary here. Perhaps, move it to _proposeUsageCost()?

maxsam4 (commented):
`changeCostAndType` allows owner to change `isCostInPoly` without any cooldown. Since 1 POLY != 1 USD, this can be used to arbitrarily increase (or decrease) usage cost. There should be a 24 hour CD for changing this as well.

maxsam4:
@satyamakgec I think some merge went wrong somewhere. I am seeing all kinds of changes here.

maxsam4 (commented):
The main changes look good. Please clean up the extra changes before merging.

pabloruiz55:
#837 has been merged. Can the PR be closed? @satyamakgec

satyamakgec:
Yes @pabloruiz55 It can be closed

PR #833, opened 2019-10-16
MATM: retrieve original allowance

ross-rosario:
- `ManualApproval` struct now contains:
  - `allowance`, which is the original supplied (or modified) allowance.
  - `remainingAllowance` which keeps track of ongoing allowance.
- Added a few tests to MATM to ensure > 95% coverage.


ross-rosario:
Ouch! I should've spun this branch from master, as we need that change in the frontend asap.

PR #834, opened 2019-10-16
MATM:enable retrieving original allowance

ross-rosario:
- `ManualApproval` struct now contains:
  - `allowance`, which is the original supplied (or modified) allowance.
  - `remainingAllowance` which keeps track of ongoing allowance.
- Added a few tests to MATM to ensure > 95% coverage.

### DIFF
--- a/contracts/modules/TransferManager/MATM/ManualApprovalTransferManager.sol
+++ b/contracts/modules/TransferManager/MATM/ManualApprovalTransferManager.sol
@@ -231,7 +231,7 @@ contract ManualApprovalTransferManager is ManualApprovalTransferManagerStorage, 
                     allowance = allowance - _changeInAllowance;
                 }
             }
-            approval.approvedAllowance = allowance;
+            approval.initialAllowance = allowance;
             approval.allowance = allowance;
         }
         // Greedy storage technique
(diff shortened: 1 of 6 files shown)


ross-rosario:
@pabloruiz55 @satyamakgec I renamed allowance variables according to your suggestion:
- `allowance` is back to keeping track of the remaining allowance.
- `approvedAllowance` stores the original allowance. I avoided calling it `original`Allowance because the word _original_ (to me) means the allowance value before the approval got _modified_ via `modifyManualApproval`, for example.

contracts/modules/TransferManager/MATM/ManualApprovalTransferManager.sol
```diff
@@ -230,6 +231,7 @@ contract ManualApprovalTransferManager is ManualApprovalTransferManagerStorage,
                     allowance = allowance - _changeInAllowance;
                 }
             }
+            approval.approvedAllowance = allowance;
```
satyamakgec: Sorry to re-iterate @remon-nashid but `initialAllowance` will be the better name for this

satyamakgec: And one more thing, `approval.approvedAllowance` value should be changed as per the increase or decrease i.e
```
       if increase 
       approvedAllowance =  approvedAllowance + changeInAllowance
       if decrease
       approvedAllowance =  approvedAllowance - changeInAllowance
```

satyamakgec (changes requested):

contracts/modules/TransferManager/MATM/ManualApprovalTransferManager.sol
ross-rosario: Done

satyamakgec (approved):

ross-rosario:
@VictorVicente we're keeping this PR open so that you utilize `initialAllowace` field in the CLI. Basically, it's a new member variable we added to `ManualApproval` struct, in order to keep a record of the initial allowance, versus the ongoing allowance tracked by `allowance` member variable.

PR #835, opened 2019-10-21
Custom oracles

satyamakgec:


contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -195,30 +199,58 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
...
+            denominatedCurrency = bytes32("USD");
+            delete customOracles[bytes32("ETH")][denominatedCurrency];
+            delete customOracles[bytes32("POLY")][denominatedCurrency];
+            oracleKeys[bytes32("ETH")][denominatedCurrency] = ETH_ORACLE;
+            oracleKeys[bytes32("POLY")][denominatedCurrency] = POLY_ORACLE;
+            emit SetOracles(denominatedCurrency, false);
+        }
+        else {
+            require(_denominatedCurrencySymbol != bytes32(0), "Invalid currency");
+            require(_customOracleAddresses.length == 2, "Invalid no. of oracles");
```
maxsam4, line 247: I think we should allow issuers to supply only ETH oracle if they are raising in ETH only.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -759,7 +803,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
      * @return address[] usd tokens
      */
     function getUsdTokens() external view returns (IERC20[] memory) {
```
maxsam4, line 816: Should this function name change as well?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
maxsam4: If they are only raising in stable coin, we shouldn't require them to enter two oracle addresses

maxsam4:
More tests for increased coverage will be appreciated as well :)

maxsam4 (approved):

PR #839, opened 2019-10-29
Coverage: additional tests for LTM

ross-rosario:
@satyamakgec besides the additional tests, I added a require statement as shown here https://github.com/PolymathNetwork/polymath-core/pull/839/files#diff-22153eb4dc048750acc26f3782d0cfdbR545 because I found that the following require will revert, without a reason, if user has no lockups at all
```
require(
            userToLockups[_userAddress][userToLockupIndex[_userAddress][_lockupName]] == _lockupName,
            "User not in lockup"
        );
```


satyamakgec (approved):
Make sense for the require statement.

PR #843, opened 2019-10-31
[WIP]:POC to support MakerDao oracle v2

satyamakgec:


contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+        currencyAddress = _currencyAddress;
+        currencySymbol = _currencySymbol;
+        polymathRegistry = IPolymathRegistry(_polymathRegistry);
+    }
+
+    /**
+      * @notice Updates medianizer address
+      * @param _medianizer Address of Maker medianizer
+      */
+    function changeMedianier(address _medianizer) public onlyOwner {
```
pabloruiz55: typo: changeMedianiZer

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+        }
+        (bytes32 price, bool valid) = medianizer.peek();
+        require(valid, "MakerDAO Oracle returning invalid value");
+        return uint256(price);
+    }
+
+    // This function allows following contracts to read data from the oracles
+    // - STR Address
+    // - Any valid module of polymath ecosystem
+    function _isCallerAllowed(address _caller) view internal {
```
adamdossa: very minor, but I'd rename to _checkCallerAllowed if we are reverting on failure inside the function (rather than returning a bool)

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+        if (_caller != polymathRegistry.getAddress(STR)) {
+            address _securityToken = IModule(_caller).securityToken();
+            // Validate the securityToken address
+            // securityToken should be a part of the Polymath ecosystem
+            require(_securityToken != address(0), "Invalid address");
+            require(ISecurityTokenRegistry(polymathRegistry.getAddress(STR)).isSecurityToken(_securityToken), "Invalid securityToken");
```
adamdossa: Should we cache `polymathRegistry.getAddress(STR)`?

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+        // All MakerDAO oracles are denominated in USD
+        return bytes32("USD");
+    }
+
+    /**
+    * @notice Returns price - should throw if not valid
+    */
+    function getPrice() external returns(uint256) {
+        _isCallerAllowed(msg.sender);
+        if (manualOverride) {
```
adamdossa, line 78: Maybe always allow users to get the manualOverride price even if they are not a module?

contracts/modules/STO/USDTiered/USDTieredSTO.sol
```diff
@@ -698,6 +698,7 @@ contract USDTieredSTO is USDTieredSTOStorage, STO {
         return (tiers[tiers.length - 1].totalTokensSoldInTier == tiers[tiers.length - 1].tokenTotal);
     }
 
+    // CONCERN: This function should be removed or change there access modifier to internal in the favor of MakerDao Oracles V2
```
adamdossa: Is there something clever we can do where the function will always revert if called on-chain, but can still be called off-chain as a view function? This would mean that no-one can use the USDTieredSTO as a way to bypass Maker's whitelist, whilst still allowing the UI etc. to access this rate.

contracts/oracles/MakerDAOOracleV2.sol
satyamakgec: Yeah it can be done

contracts/modules/STO/USDTiered/USDTieredSTO.sol
maxsam4: One way will be to read the makerdao oracle's value offchain and then send it as a parameter to this function.
Onchain, we'll need a separate internal function that fetches the value from the oracle.

btw, Why do we even want to limit this function from being called onchain? Does makerdao now charge a fee based on the number of calls made? In their medianizer contract, it's a  simple whitelist check but I am not sure what the offchain licensing terms are.

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+    /**
+      * @notice Creates a new Maker based oracle
+      * @param _medianizer Address of Maker medianizer
+      * @param _currencyAddress Address of currency (0x0 for ETH)
+      * @param _currencySymbol Symbol of currency
+      * @param _polymathRegistry Address of the Polymath Registry
+      */
+    constructor(address _medianizer, address _currencyAddress, bytes32 _currencySymbol, address _polymathRegistry) public {
+        medianizer = IMedianizer(_medianizer);
+        currencyAddress = _currencyAddress;
```
maxsam4, line 35: Why do we need this?

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+        currencySymbol = _currencySymbol;
```
maxsam4, line 36: Why do we need this?

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+    * @notice Returns symbol of oracle currency (0x0 for ETH)
+    */
+    function getCurrencySymbol() external view returns(bytes32) {
+        return currencySymbol;
+    }
+
+    /**
+    * @notice Returns denomination of price
+    */
+    function getCurrencyDenominated() external view returns(bytes32) {
```
maxsam4, line 69: Why do we need this?

contracts/oracles/MakerDAOOracleV2.sol
```diff
@@ -0,0 +1,126 @@
...
+            (,moduleAddress,,isArchived,,) = ISecurityToken(_securityToken).getModule(_caller);
+            require(moduleAddress == _caller && !isArchived, "Invalid msg.sender");
+        }
+    }
+
+    /**
+      * @notice Set a manual price. NA - this will only be used if manualOverride == true
+      * @param _price Price to set
+      */
+    function setManualPrice(uint256 _price) public onlyOwner {
```
maxsam4, line 110: Perhaps we should have something like `setManualOverrideWithPrice` that sets both the manual override bool and manual price. It will eliminate the need for making two calls to set manual override.

contracts/modules/STO/USDTiered/USDTieredSTO.sol
satyamakgec: I think the clever way is to use msg.sender as 0x0

contracts/oracles/MakerDAOOracleV2.sol
satyamakgec: To know what is the currency address for which we created a wrapper oracle

contracts/oracles/MakerDAOOracleV2.sol
satyamakgec: same as above change

contracts/oracles/MakerDAOOracleV2.sol
satyamakgec: We need this to know the denominated currency, Down the line, we may introduce different fiat currency pegged oracles

contracts/oracles/MakerDAOOracleV2.sol
maxsam4: I don't think we'll ever be needing these variables on-chain. It just doesn't make sense for a contract to call another contract and ask "hey, which oracle are you?". The caller contract should always exactly know which oracle it is calling. 

anyway, perhaps we should make these constants rather than variables? That will make it cheaper and we don't have a function to change these anyway.

PR #857, opened 2019-11-19
SC-19/improve schedule checkpoint module

satyamakgec:


contracts/modules/Checkpoint/Automation/ScheduleCheckpoint.sol
```diff
@@ -0,0 +1,269 @@
...
+     * @param _frequency How frequent checkpoint will being created
+     * @param _frequencyUnit Unit of frequency i.e If issuer puts _frequency = 10
+     * & frequency unit is DAYS then it means every 10 day frequency new checkpoint will be created
+     */
+    function addSchedule(bytes32 _name, uint256 _startTime, uint256 _endTime, uint256 _frequency, FrequencyUnit _frequencyUnit) withPerm(OPERATOR) external {
+        require(_name != bytes32(0), "Empty name");
+        require(_startTime >= now, "Start time must be in the future");
+        require(schedules[_name].name == bytes32(0), "Name already in use");
+        uint256 endTime = _endTime;
+        if (_endTime <= _startTime)
```
adamdossa, line 49: Shouldn't we require that either _endTime == 0, or _endTime > _startTime

contracts/modules/Checkpoint/Automation/ScheduleCheckpoint.sol
```diff
@@ -0,0 +1,269 @@
...
+     * @param _name name of the new schedule (must be unused)
+     * @param _startTime start time of the schedule (first checkpoint)
+     * @param _endTime End time of the schedule
+     * @param _frequency How frequent checkpoint will being created
+     * @param _frequencyUnit Unit of frequency i.e If issuer puts _frequency = 10
+     * & frequency unit is DAYS then it means every 10 day frequency new checkpoint will be created
+     */
+    function addSchedule(bytes32 _name, uint256 _startTime, uint256 _endTime, uint256 _frequency, FrequencyUnit _frequencyUnit) withPerm(OPERATOR) external {
+        require(_name != bytes32(0), "Empty name");
+        require(_startTime >= now, "Start time must be in the future");
```
adamdossa: I think this should be strictly greater than, not greater than or equal? If it is equal to now, then the schedule will not actually be processed in this block.

contracts/modules/Checkpoint/Automation/ScheduleCheckpoint.sol
```diff
@@ -0,0 +1,269 @@
...
+     * @param _newEndTime New end time of the schedule
+     */
+    function modifyScheduleEndTime(bytes32 _name, uint256 _newEndTime) withPerm(OPERATOR) external {
+        Schedule memory _schedule = schedules[_name];
+        require(_schedule.name != bytes32(0), "Invalid name");
+        if (_schedule.endTime > 0)
+            require(_schedule.endTime > now, "Schedule already ended");
+        if (_newEndTime > 0)
+            require(_newEndTime > now, "Invalid end time");
+        emit ModifyScheduleEndTime(_name, _schedule.endTime, _newEndTime);
```
adamdossa, line 93: Should we also check that _newEndTime > _schedule.startTime?

contracts/modules/Checkpoint/Automation/ScheduleCheckpoint.sol
```diff
@@ -0,0 +1,269 @@
...
+        if (_endTime <= _startTime)
+            endTime = uint256(0);
+        schedules[_name].name = _name;
+        schedules[_name].startTime = _startTime;
+        schedules[_name].endTime = endTime;
+        schedules[_name].nextCheckPointCreatedAt = _startTime;
+        schedules[_name].frequency = _frequency;
+        schedules[_name].frequencyUnit = _frequencyUnit;
+        schedules[_name].index = names.length;
+        names.push(_name);
```
adamdossa, line 58: Do you think we should limit the number of schedules (given the gas costs etc.)? I suspect most people will only need one or two schedules so we could limit this without impacting the functionality for most people, and guarding against transfers becoming too expensive.

contracts/modules/Checkpoint/Automation/ScheduleCheckpointStorage.sol
```diff
@@ -0,0 +1,27 @@
...
+    enum FrequencyUnit { SECONDS, DAYS, WEEKS, MONTHS, QUATER, YEARS }
+    
+    bytes32 constant OPERATOR = "OPERATOR";
+
+    struct Schedule {
+        uint256 index;
+        bytes32 name;
+        uint256 startTime;
+        uint256 endTime;
+        uint256 nextCheckPointCreatedAt;
```
maxsam4: `nextCheckPointCreatedAt` is confuing. Rename to `createNextCheckpoint` maybe?
Then add a code comment in the update function that since checkpoint has been created, this is now being treated as `lastCheckpointCreatedAt`.

maxsam4 (commented):
Apart from what Adam mentioned, it looks fine overall. 

However, I am not sold that we should be doing scheduling onchain. It doesn't seem like worth the increased transfer costs. I'd suggest that scheduling should be done with the help of oracles or just running a service offchain that calls the required function after X days or whatever.

contracts/modules/Checkpoint/Automation/ScheduleCheckpoint.sol
satyamakgec: endTime is not a mandatory parameter, So I am not restricting an issuer about the endTime value. If they put anything that is below or equal to startTime it will make endTime = 0.

ross-rosario:
Added schedules limit validation (<= 10) per discussion with @satyamakgec.

maxsam4 (approved):

PR #860, opened 2019-11-26
Use ganache cli with Istanbul hard fork

satyamakgec:


scripts/test.sh
```diff
@@ -59,7 +59,7 @@ start_testrpc() {
   if [ "$COVERAGE" = true ]; then
     node --max-old-space-size=8192 node_modules/.bin/testrpc-sc --gasLimit 0xfffffffff --port "$testrpc_port" "${accounts[@]}" > /dev/null &
   else
-    node_modules/.bin/ganache-cli --gasLimit 8000000 "${accounts[@]}" > /dev/null &
+    node_modules/.bin/ganache-cli --hardfork istanbul --gasLimit 8000000 "${accounts[@]}" > /dev/null &
```
ross-rosario: Shall we change the gasLimit here to 10,000,000?

ross-rosario: Same for package.json

satyamakgec: I want to keep as it is so we can get an alarm

maxsam4: I'd vote to set it to at least 9.5M. This way, we have enough headroom and we can recommend higher values for batch functions. I don't have a strong opinion about this though. You need to set a higher gas price for heavier tx so I am usually not a fan of big tx.

satyamakgec:
@remon-nashid @maxsam4 could someone approve this?

ross-rosario:
I was going to, but then I stumbled upon this https://github.com/trufflesuite/ganache-cli/issues/701. @satyamakgec shall we wait until it's resolved?

maxsam4:
I don't think we should be waiting for the ganache fix. Gas calculation in ganache is broken with tons of edge cases. It should be treated as a rough estimate and not an exact gas cost.

maxsam4 (approved):

maxsam4:
I'd like to add that although this PR changes Ganache runtime to Istanbul, this does not upgrade solidity evm target. As a result, contracts will be compiled with older EVM version as target but executed on newer EVM (exactly how our mainnet contracts will work after the HF).

Before the next major release, we might want to consider upgrading our solidity evm target.

satyamakgec:
Yeah @maxsam4 that is the plan.

davidmurdoch:
@maxsam4 What edge cases for gas calculations are you referring to? Are they specific to istanbul? /cc @nicholasjpaterno

maxsam4:
@davidmurdoch last time I did profiling of our functions in Ganache was in December last year so a lot probably has changed since then. Some edge cases I remember were:
1) When a contract emits an event then does a delegate call (proxy) to a contract that emits an event and then it does a normal call to a contract that emits an event. Gas was charged for only some events.
2) There were some inconsistencies around packing. Let's say a contract updates 4 tightly packed state variables from 0 -> 1, then gas-charged should be 20k + 5k +5k +5k but Ganache was charging 20k + 20k + 20k +20k as if every variable consumed a new slot.

Since I have not done these tests recently, I know of no Istanbul related gas calculation edge cases.

